PR #51, opened 2026-03-10
Update the Anthropic Models

carlospedreira:
## Summary

Updates the Claude model picker and Anthropic runtime defaults to use Claude Code’s rolling model aliases (`opus`, `sonnet`, `haiku`) instead of hardcoded versioned model IDs.

## What changed

- Updated the Claude provider mapping so `"Opus"`, `"Sonnet"`, and `"Haiku"` resolve to `opus`, `sonnet`, and `haiku`
- Updated `ProviderCLI` wording to reflect that model resolution now returns the provider’s configured selector rather than a pinned versioned ID
- Updated `ClaudeCLI` defaults, examples, and model listings to use rolling aliases instead of hardcoded `claude-*` model names
- Updated Claude-related tests to expect alias-based resolution (`Opus` -> `opus`, E2E Haiku -> `haiku`)

## Why

This makes the model picker follow Anthropic’s latest model for each family automatically, instead of requiring repo changes every time Anthropic publishes a new pinned model version.


carlospedreira:
@andresharpe the request to dynamically load this from the actual providers is a bit more difficult than anticipated - I would merge this one and put the dynamic model discovery in a separate issue. Do you agree?

PR #110, opened 2026-04-03
feat: open kickstart modal for workflows requiring user input

### LINKED ISSUE #105: Jira workflow does not show kickstart dialog
- Create a new folder
- dotbot init -Workflow kickstart-via-jira
- ./.bot/go.ps1
- Click Run on workflow
- It should display a Kickstart Dialog

Instead it creates tasks with no input and it fails to continue since there is no data on what to do

bbolek-ap:
Fixes #105 

This pull request updates the workflow UI to better handle workflows that require user input forms before running. The main improvement is that when a workflow defines a form, clicking "Run" will now open a modal dialog for user input instead of immediately starting the workflow. This provides a more intuitive experience for workflows that need additional information from the user.

**UI and Workflow Execution Improvements:**

* Added a new `has_form` property to the workflow status object in `server.ps1` to indicate if a workflow requires a form before running.
* Updated the `runWorkflow` function and all its call sites in the UI (`controls.js`, `kickstart.js`, `workflow.js`) to accept a `hasForm` parameter and open the kickstart modal if the workflow requires user input, instead of running immediately. [[1]](diffhunk://#diff-1f1cb32aebcd48640eaff83d7b8fcdb14e3a50d5ac0f7431724d900cafd9f4c0L463-R482) [[2]](diffhunk://#diff-179d2bd64f80e7a36a4a0a12f77b0145427a2880ff87d254d6e38b9f899f05eaL323-R323) [[3]](diffhunk://#diff-98e71c91f8e6885c6d7ba2f2c9572a4678cc0837be7ca3aa1aec5e355526c99dL655-R655)

### DIFF
--- a/workflows/default/systems/ui/static/modules/kickstart.js
+++ b/workflows/default/systems/ui/static/modules/kickstart.js
@@ -226,7 +226,7 @@ function renderKickstartCTA(container) {
                 <div class="kickstart-title">${escapeHtml(title)}</div>
                 <div class="kickstart-description">${escapeHtml(desc)}</div>
                 ${phaseNames ? `<div class="kickstart-phase-inline">${phaseNames}</div>` : ''}
-                <button class="kickstart-btn" onclick="openKickstartModal()" style="margin-top: 1.5rem">${escapeHtml(buttonText)}</button>
+                <button class="kickstart-btn" onclick="openKickstartModal(currentWorkflowName)" style="margin-top: 1.5rem">${escapeHtml(buttonText)}</button>
             </div>
         `;
         return;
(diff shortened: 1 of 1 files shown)


andresharpe (dismissed):
Thanks for tackling #105 — the approach of checking `has_form` before running is the right direction. A couple of changes needed before this can merge:

### 1. Pass the workflow name through to `openKickstartModal()`

Currently `openKickstartModal()` is called with no arguments, so the modal always renders the default/active workflow's form. In a multi-workflow setup (e.g., Jira + default), clicking Run on the Jira workflow would show the default workflow's kickstart form — wrong context, wrong fields.

**`controls.js`** — pass `name` into the modal:
```js
if (hasForm) {
    if (typeof openKickstartModal === 'function') {
        openKickstartModal(name);  // ← so the modal knows which workflow to configure
    }
    ...
}
```

`openKickstartModal()` (in `kickstart.js`) will need to accept a workflow name parameter and use it when rendering the form and submitting the kickstart.

### 2. Consistent `hasForm` detection in `kickstart.js`

In `kickstart.js:323` you use `kickstartDialog != null` to determine `hasForm`, but the other two call sites (`controls.js`, `workflow.js`) use `wf.has_form` from the server API. These could diverge — e.g., a workflow without a form in a session where a global kickstart dialog exists.

For consistency, the workflow card grid in `kickstart.js` should also use the server-provided `has_form` property from the workflow status data, rather than the client-side `kickstartDialog` variable.

Issue #103, opened 2026-04-01
Hide global analyze-execute controls, use per-workflow controls


andresharpe:
## Summary

Per-workflow start/stop controls exist (`controls.js` lines 417-540) but the generic global "Execute Tasks" buttons are still visible in `index.html` (lines 96-105, id=`generic-workflow-control`).

## What to Do

- Hide or remove the global start-workflow/stop-workflow/kill-workflow buttons
- Per-workflow controls already implemented with LED status indicators — these should be the only way to control workflows

## Related

- See `ideas/roadmap/erol-05-hide-analyze-execute-controls.md`

kabaogluemre:
@andresharpe  I would like to take over this.

PR #121, opened 2026-04-06
Hide global analyze-execute controls, use per-workflow controls #103

kabaogluemre:
**Summary**
Hide the generic "Execute Tasks" global control (#generic-workflow-control) now that per-workflow Start/Stop controls exist
Remove JS fallback that re-showed the global control when no workflows were installed
Also, update stale JSDoc comment to reflect the new behavior

### DIFF
--- a/workflows/default/systems/ui/static/index.html
+++ b/workflows/default/systems/ui/static/index.html
@@ -94,7 +94,7 @@
                                 <!-- Dynamic workflow controls (populated by JS when workflows are installed) -->
                                 <div id="workflow-controls-container"></div>
                                 <!-- Execute Tasks: launches workflow process to execute pending tasks -->
-                                <div class="process-control-row" id="generic-workflow-control" style="display: none">
+                                <div class="process-control-row hidden" id="generic-workflow-control">
                                     <div class="process-control-header">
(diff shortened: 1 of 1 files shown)


andresharpe (commented):
## Review

**Overall:** Clean, focused PR that correctly completes issue #103. The change is small and does exactly what's needed — no existing tests will break.

### Copilot's inline style comment — agree

The suggestion to use `class="process-control-row hidden"` instead of `style="display: none"` is a good one. The `.hidden` class is already defined in `base.css:92` and used throughout the codebase (sidebar, modals, action widgets). Using it here keeps styling centralized and consistent.

### Edge case: no workflows installed

With the fallback removed from `renderWorkflowControls`, when no workflows are installed, **both** the generic control and per-workflow controls will be absent — the controls area will be completely empty. This is probably fine (if there are no workflows, there's nothing to execute), but worth confirming that's the intended UX. A subtle "No workflows installed" placeholder could be considered in a follow-up.

### Test coverage

No automated tests cover `renderWorkflowControls` or the `#generic-workflow-control` element (the project has no JS unit tests — this is a pre-existing gap, not introduced by this PR). No existing tests will break from these changes.

PR #122, opened 2026-04-06
Fix blocking issues from structured logging refactor

kabaogluemre:
## Fix blocking issues from structured logging refactor

The recent structured logging commit introduced several `Set-StrictMode` violations that broke both the Web UI startup and kickstart/task launches.

### Changes

**`go.ps1` — `Write-Status` not recognized**
- The Platform-Functions module (which provides `Write-Status`) was imported *after* the first `Write-Status` call. Moved the `Import-Module Platform-Functions.psm1` block up before line 55, right after logging is initialized, so `Write-Status` is available when first called.

**`launch-process.ps1` — `$uiSettings` uninitialized**
- `$uiSettings` was only assigned inside `if (Test-Path $uiSettingsPath)` but referenced unconditionally later. Added `$uiSettings = $null` initialization before the conditional block.

**`launch-process.ps1` — `$procId` and `$lockKey` uninitialized in crash trap**
- The `trap {}` block referenced `$procId`, `$processData`, and `$lockKey`, but these were assigned *after* the trap registration. If the script crashed before reaching those assignments, strict mode would throw.
- Moved `$procId` and `$lockKey` initialization above the trap block.
- Added `Test-Path variable:` guards inside the trap for `$procId`, `$processData`, and `$lockKey` so the trap degrades gracefully even if variables are partially initialized.
- Wrapped `Remove-ProcessLock` in its own guarded block so lock cleanup doesn't fail if `$lockKey` was never set.

### DIFF
--- a/workflows/default/systems/runtime/launch-process.ps1
+++ b/workflows/default/systems/runtime/launch-process.ps1
@@ -258,7 +258,7 @@ trap {
         $processData.failed_at = (Get-Date).ToUniversalTime().ToString("o")
         $processData.error = "Unexpected termination: $($_.Exception.Message)"
         try { Write-ProcessFile -Id $procId -Data $processData } catch { Write-BotLog -Level Debug -Message "Non-critical operation failed" -Exception $_ }
-        try { Write-ProcessActivity -Id $procId -ActivityType "text" -Message "Process terminated unexpectedly: $($_.Exception.Message)" } catch { Write-BotLog -Level Warn -Message "Failed to read process data" -Exception $_ }
+        try { Write-ProcessActivity -Id $procId -ActivityType "text" -Message "Process terminated unexpectedly: $($_.Exception.Message)" } catch { Write-BotLog -Level Warn -Message "Failed to write process activity" -Exception $_ }
     }
     if (Test-Path variable:lockKey) {
         try { Remove-ProcessLock -LockType $lockKey } catch { Write-BotLog -Level Debug -Message "Logging operation failed" -Exception $_ }


kabaogluemre:
@andresharpe , by any chance can you check my changes, since these are blocking issues.

andresharpe (commented):
## `go.ps1` — Platform-Functions is the wrong module to move here

Good catch that `Write-Status` is called on line 53 before any module providing it is imported — that's a real bug. However, this change imports the wrong module, which would actually make things worse.

**The problem:** There are two different `Write-Status` functions with incompatible signatures:

| Module | Signature | Supports `-Type`? |
|--------|-----------|-------------------|
| `Platform-Functions.psm1` | `param([string]$Message)` | No |
| `DotBotTheme.psm1` | `param([string]$Message, [string]$Type)` | Yes |

Every call site in `go.ps1` uses the `-Type` parameter:
```powershell
Write-Status "  Starting .bot UI..." -Type Info        # line 53
Write-Status "  Server already running..." -Type Success  # line 74
Write-Status "  Starting UI server..." -Type Info      # line 108
```

If `Platform-Functions.psm1` is imported early (as this PR does), those calls would resolve to the version without `-Type` and throw:
> *"A parameter cannot be found that matches parameter name 'Type'"*

**The correct fix** is to import `DotBotTheme.psm1` before line 53 — that's the module with the `Write-Status` signature that `go.ps1` actually uses. Something like:

```powershell
Import-Module "$PSScriptRoot\systems\runtime\modules\DotBotLog.psm1" -Force -DisableNameChecking
Initialize-DotBotLog -LogDir $logsDir -ControlDir $controlDir -ProjectRoot (Split-Path $BotDir -Parent)

# Import theme module (provides Write-Status with -Type parameter)
Import-Module "$PSScriptRoot\systems\runtime\modules\DotBotTheme.psm1" -Force -DisableNameChecking
Initialize-DotBotTheme

Write-BotLog -Level Info -Message "go.ps1 launched. BotDir=$BotDir"
Write-Status "  Starting .bot UI..." -Type Info
```

The reason this works on most machines today is likely that `DotBotTheme.psm1` is already loaded in the session from a prior `dotbot` operation or global install. On a clean environment (like yours), nothing has loaded it yet, so `Write-Status` is genuinely undefined at line 53.

---

The `launch-process.ps1` changes all look correct — the `$uiSettings = $null` init, the trap variable hoisting, and the `Test-Path variable:` guards are all good fixes for real StrictMode 1.0 issues. 👍

PR #260, opened 2026-04-14
feat: framework file protection, prevent agents from modifying .bot/…

### LINKED ISSUE #173: Framework file protection: prevent agents from modifying .bot/ framework files
## Problem

AI coding agents (Claude, Codex, Gemini) working in dotbot target projects can accidentally modify framework-owned files in `.bot/` that should only be changed via `dotbot init --force`. There is currently no protection mechanism — no manifest, no pre-commit guard, and no agent instructions warning against it.

This is a real risk because agents see `.bot/` files in their context and may "helpfully" edit them to fix perceived issues, add features, or refactor code — corrupting the framework installation.

### Layer 1: Manifest + Integrity Hashes

**New file:** `scripts/Manifest.psm1`

A shared PowerShell module with two functions:

- `New-DotbotManifest` — walks all framework-owned files under `.bot/`, computes SHA256 hashes via `Get-FileHash`, writes `.bot/.manifest.json` (committed to git)
- `Test-DotbotManifest` — verifies file hashes against manifest, reports tampered/missing files

**Manifest format** (`.bot/.manifest.json`):
```json
{
  "version": 1,
  "generated_at": "2026-04-08T12:00:00Z",
  "generator": "dotbot/x.x",
  "protected_paths": [
    "systems/", "hooks/", "recipes/",
    "settings/providers/", "settings/settings.default.json",
    "settings/theme.default.json",
    "go.ps1", "init.ps1", "README.md", ".gitignore",
    "workflow.yaml", ".manifest.json"
  ],
  "user_paths": ["workspace/", ".control/", "profile/"],
  "files": {
    "systems/mcp/dotbot-mcp.ps1": { "sha256": "a1b2c3...", "size": 12345 },
    "hooks/verify/00-privacy-scan.ps1": { "sha256": "d4e5f6...", "size": 6338 }
  }
}
```

**Integration:** Called at the end of `scripts/init-project.ps1` after pre-commit hook installation (after line 1092), before the initial commit. Also regenerated during `dotbot init --force` after all copy operations complete.

**Design decision:** Manifest is **committed** (not gitignored) so drift detection works on any clone or CI run without needing to run init first. The manifest file itself is in the protected paths list, so it cannot be tampered with by agents.

---

### Layer 2: Pre-commit Hook (Pure Bash, ~10ms)

**Modified file:** `scripts/init-project.ps1` (lines 1059-1085 hook template)

Add a framework file protection check to the existing generated `.git/hooks/pre-commit` bash script:

```bash
# --- dotbot framework file protection ---
PROTECTED_FILES=""
for staged in $(git diff --cached --name-only); do
  case "$staged" in
    .bot/systems/*|.bot/hooks/*|.bot/recipes/*|\
    .bot/settings/providers/*|.bot/settings/settings.default.json|\
    .bot/settings/theme.default.json|\
    .bot/go.ps1|.bot/init.ps1|.bot/README.md|.bot/.gitignore|\
    .bot/workflow.yaml|.bot/.manifest.json)
      PROTECTED_FILES="$PROTECTED_FILES  $staged\n"
      ;;
  esac
done
if [ -n "$PROTECTED_FILES" ]; then
  echo ""
  echo "ERROR: Framework-owned files cannot be modified directly."
  echo ""
  echo "Blocked files:"
  echo -e "$PROTECTED_FILES"
  echo "To update framework files: dotbot init --force"
  echo "To discard changes: git checkout -- <file>"
  exit 1
fi
```

**Why pure bash:** A pwsh invocation adds 500-1500ms startup overhead. The bash `case` statement against `git diff --cached --name-only` runs in ~10ms. Protected paths are hardcoded in the template and regenerated on `dotbot init`.

---

### Layer 3: Verification Hook (Deep Integrity Check at Task Completion)

**New file:** `workflows/default/hooks/verify/04-framework-integrity.ps1`

Follows the existing verify hook pattern (JSON output protocol, `$TaskId`/`$Category` params):

- Loads `.bot/.manifest.json`
- For every file in `manifest.files`: verifies SHA256 hash matches
- Reports tampered files, missing files
- Output: `{ success, script, message, details: { files_checked, tampered, missing }, failures }`

**Registration in** `workflows/default/hooks/verify/config.json`:
```json
{
  "name": "04-framework-integrity.ps1",
  "description": "Verify framework-owned files have not been modified",
  "required": true,
  "core": true,
  "timeout_seconds": 30
}
```

This runs at `task-mark-done` time — catches modifications that somehow bypassed the pre-commit hook (e.g., agent using `--no-verify`, or direct file writes without committing).

---

### Layer 4: Agent Instructions in Project CLAUDE.md

**Modified file:** `workflows/default/init.ps1`

After the existing agent/skill copy loop, append a framework protection section to the target project's `CLAUDE.md` (and equivalent files for other providers). Uses an idempotent marker (`<!-- dotbot:framework-protection -->`) so re-running init does not duplicate the section.

```markdown
## dotbot Framework Files (READ-ONLY)

The `.bot/` directory contains framework files managed by dotbot.

### Protected (DO NOT MODIFY):
- `.bot/systems/` — MCP server, runtime, web UI
- `.bot/hooks/` — verification and lifecycle hooks
- `.bot/recipes/` — agents, skills, prompts
- `.bot/settings/providers/`, `.bot/settings/*.default.json`
- `.bot/go.ps1`, `.bot/init.ps1`, `.bot/workflow.yaml`, `.bot/.manifest.json`

### Safe to edit:
- `.bot/workspace/` — tasks, sessions, decisions, plans, feedback, reports

### To update framework files: run `dotbot init --force`
```

---

### Layer 5: File Headers (Key Files Only)

Add `DO NOT MODIFY` banners to ~15 high-value framework files that agents are most likely to touch:

**PowerShell format:**
```powershell
# ═══════════════════════════════════════════════════════════════
# FRAMEWORK FILE — DO NOT MODIFY IN TARGET PROJECTS
# Managed by dotbot. Overwritten on 'dotbot init --force'.
# ═══════════════════════════════════════════════════════════════
```

**Target files:**
- `workflows/default/go.ps1`
- `workflows/default/init.ps1`
- `workflows/default/systems/mcp/dotbot-mcp.ps1`
- `workflows/default/hooks/verify/00-privacy-scan.ps1`
- `workflows/default/hooks/verify/01-git-clean.ps1`
- `workflows/default/hooks/verify/02-git-pushed.ps1`
- `workflows/default/hooks/verify/03-check-md-refs.ps1`
- `workflows/default/hooks/verify/04-framework-integrity.ps1` (new)
- `workflows/default/hooks/scripts/commit-bot-state.ps1`
- `workflows/default/hooks/scripts/steering.ps1`
- `workflows/default/hooks/dev/Start-Dev.ps1`
- `workflows/default/hooks/dev/Stop-Dev.ps1`
- `workflows/default/workflow.yaml` (YAML comment)
- `workflows/default/recipes/agents/*/AGENT.md` (HTML comment)

---

### Layer 6: Structure Tests

**Modified file:** `tests/Test-Structure.ps1`

Using the project's custom test framework (`Assert-True` / `Write-TestResult` from `Test-Helpers.psm1`). New `FRAMEWORK FILE PROTECTION` section:

1. Manifest module exists and is importable
2. Manifest generates valid JSON with required keys (`version`, `protected_paths`, `user_paths`, `files`)
3. Manifest file entries have `sha256` and `size` fields
4. Hash round-trip is correct (independently hash `go.ps1`, compare to manifest)
5. Protected paths cover core directories (`systems/`, `hooks/`, `recipes/`)
6. Framework integrity hook is registered in `config.json`
7. Key framework files have `DO NOT MODIFY` header

---

## File Classification

| Path (relative to `.bot/`) | Owner | Committed | Modifiable |
|---|---|---|---|
| `systems/` | Framework | Yes | No |
| `hooks/` | Framework | Yes | No |
| `recipes/` | Framework | Yes | No |
| `settings/providers/` | Framework | Yes | No |
| `settings/*.default.json` | Framework | Yes | No |
| `go.ps1`, `init.ps1`, `workflow.yaml` | Framework | Yes | No |
| `.manifest.json` | Framework | Yes | No |
| `workspace/` | User | Yes | Yes |
| `.control/` | Runtime | No (gitignored) | N/A |
| `profile/` | User | No (gitignored) | N/A |

## Files to Create
- `scripts/Manifest.psm1`
- `workflows/default/hooks/verify/04-framework-integrity.ps1`

## Files to Modify
- `scripts/init-project.ps1` — manifest generation + pre-commit hook extension
- `workflows/default/hooks/verify/config.json` — register new hook
- `workflows/default/init.ps1` — agent instructions in CLAUDE.md
- `tests/Test-Structure.ps1` — new test section
- ~15 framework files — add DO NOT MODIFY headers

## Implementation Order
1. Manifest module (`scripts/Manifest.psm1`)
2. Manifest integration in `scripts/init-project.ps1`
3. Pre-commit hook extension in `scripts/init-project.ps1`
4. Verification hook (`04-framework-integrity.ps1` + `config.json`)
5. Agent instructions in `workflows/default/init.ps1`
6. File headers on ~15 key files
7. Structure tests in `tests/Test-Structure.ps1`
8. Full test suite pass

gmireles-ap:
## Linked issue

Closes #173

## Summary of changes

AI coding agents (Claude, Codex, Gemini) working in dotbot target projects can accidentally modify framework-owned files in `.bot/`. This PR implements the six-layer defense strategy proposed in #173:

**Layer 1: Manifest + Integrity Hashes** — New `scripts/Manifest.psm1` module generates `.bot/.manifest.json` with SHA256 hashes of every framework file, committed to git for cross-clone verification. `New-DotbotManifest` generates the manifest; `Test-DotbotManifest` verifies hashes and reports tampered, missing, or added files.

**Layer 2: Pre-commit Hook (~10ms, pure bash)** — Blocks staging of protected files using a `case` statement matching framework paths (`.bot/systems/*`, `.bot/hooks/*`, `.bot/recipes/*`, etc.). `DOTBOT_FORCE_COMMIT=1` escape hatch for the installer. Protected paths are auto-generated from the canonical list so the hook stays in sync.

**Layer 3: Verification Hook** — New `04-framework-integrity.ps1` performs deep integrity checking, verifying SHA256 hashes against the manifest and reporting tampered or missing files. Registered in `config.json` as `required: true, core: true`. Catches tampering committed via `git commit --no-verify`.

**Layer 4: Agent Instructions** — README updated to document that `.bot/` must stay tracked in git, with clear guidance that `dotbot init --force` is required for framework updates.

**Layer 5: File Headers** — Key framework files receive "DO NOT MODIFY — Managed by dotbot. Overwritten on `dotbot init --force`" banners.

**Layer 6: Structure Tests** — Validates manifest generation, JSON validity, SHA256 hash accuracy, tamper/add/delete detection, protected path coverage, hook registration, and MCP gate wiring.

### Beyond the original spec

- **Integrity gates on MCP task transitions** — Per @ProtonPump's suggestion that "if an agent commits with --no-verify and the task never reaches mark-done, the tampering goes undetected", `task-mark-analysing` and `task-mark-in-progress` now run integrity checks before allowing state transitions. This is particularly relevant because analysis runs on main without worktree isolation.
- **Gitignore warning** — `init-project.ps1` now warns when `.bot/` is gitignored, since framework integrity checks rely on git visibility.

## Screenshots / recordings

N/A — no UI changes.

## Testing notes

```bash
# Run layers 1-3 (no credentials needed)
pwsh tests/Run-Tests.ps1
```

## Small deviations from proposal

Two places where the implementation differs from issue #173.

### 1. Manifest.psm1 lives in `workflows/default/systems/mcp/modules/` instead of `scripts/`

@ProtonPump suggested running the integrity check at task-mark-analysing (when entering a worktree), in addition to task completion. That way a `--no-verify` bypass still gets caught even if a task never reaches mark-done. This was implemented by wiring the check into the MCP task gates.

That gate runs inside the user's target project, from an MCP tool. So the hash code needs to be accessible from inside `.bot/`. Shipping Manifest.psm1 through the normal `workflows/default/* -> .bot/*` copy is the simplest path. It keeps `.bot/` self-contained and removes the sidecar `Copy-Item` block that would otherwise live in `init-project.ps1` just to deliver one file.

### 2. Hashing uses `Get-FrameworkContentHash` instead of `Get-FileHash`

`Get-FileHash` is byte-level SHA256. This was implemented differently because of the following user scenario:

1. User A on Windows runs `dotbot init` and commits `.bot/`.
2. User B on Linux clones the repo.
3. User B starts a task. The integrity gate blocks it with "framework tampered", even though no one edited a framework file.

What happens under the hood: on Windows, `init-project.ps1` uses `Set-Content` to write protected files like `.bot/settings/settings.default.json`. `Set-Content` writes CRLF bytes. The manifest records CRLF hashes. At `git add` time, git converts CRLF to LF in the index. User B clones and gets LF on disk. Hashing LF does not match the stored CRLF hash, so the gate blocks.

Reproduced this on both Windows and Linux with matching hashes, so the drift is caused by git's line-ending filter, not by the OS itself.

I first tried shipping `.gitattributes` with `* text=auto eol=lf` to keep the bytes consistent on disk. It was too fragile. It required every `Copy-Item` source to already be LF, every `Set-Content` in `init-project.ps1` to write LF on Windows (which it does not by default), and no future side-channel writes anywhere.

`Get-FrameworkContentHash` computes SHA256 over the file's content with carriage return bytes stripped out, so CRLF and LF versions of the same file produce the same hash. This sidesteps the line-ending question entirely. Real content changes are still detected (a negative-control test appends a byte and confirms the hash changes).

## Checklist

- [x] Tests added or updated
- [x] Docs updated (if behaviour changed)
- [x] Linked issue exists


andresharpe (approved):

gmireles-ap:
@andresharpe Quick design question before merging this PR.

This framework integrity feature hard-blocks when .bot/ is gitignored: MCP task gates refuse to start work, and the pre-commit hook prevents untracking .bot/ once it's tracked (git rm --cached stages deletions, hook rejects them).

Today, dotbot works fine on a single machine with .bot/ ignored. Worktrees use filesystem junctions, not git tracking. The only thing lost is state persistence across clones.

Is the vision for .bot/ to always be tracked, no opt-out? If not, should integrity gracefully disable itself when .bot/ is ignored instead of blocking?

andresharpe:
@gmireles-ap good question — worth pinning down.

The philosophy is that a repo carries both the product *and* the record of how it was built plus the decisions made along the way. `.bot/` (manifest, settings, decisions, workspace/product artifacts) is part of that institutional memory. So the vision is **always tracked, no opt-out** — and the hard-block in this PR is the correct behavior.

For existing projects that currently have `.bot/` gitignored, treat it as a one-time migration, not a supported configuration: `git rm --cached -r .bot/...`, drop the `.bot/` entry from `.gitignore`, `dotbot init --force`. If the block message could nudge users toward that sequence (a hint line when it detects the gitignore case), that would make the migration path obvious without softening the contract.

Happy to merge this as-is. Thanks for raising it before the fact.

PR #303, opened 2026-04-21
fix:Dotbot does not continue working on task #214

### LINKED ISSUE #214: Dotbot does not continue working on task
## Problem

When a workflow is started and killed mid-analysis (task in \`analysing\` status), restarting execution does not pick up the interrupted task. The task remains stuck in \`analysing/\` indefinitely.

## Root Cause

Two issues combine to create the deadlock:

### 1. Over-conservative staleness buffer in task-reset.ps1

\`Reset-AnalysingTasks\` (lines 178-280) cross-references analysing tasks against live processes. An orphaned task is only recovered if its \`updated_at\` is older than **5 minutes**:

\`\`\`powershell
# Line 252
\$stalenessThreshold = \$now.AddMinutes(-5)

# Line 267-270 — skips tasks updated less than 5 minutes ago
if (\$taskContent.updated_at) {
    if (\$updatedAt -gt \$stalenessThreshold) { continue }  # SKIPS RECENT KILLS
}
\`\`\`

When a process is killed, its task's \`updated_at\` is recent. The staleness check skips it even though:
- The owning process is confirmed dead (PID check at lines 232-239)
- The task is NOT in the live task ID set (line 264)

The 5-minute buffer was designed to avoid racing with freshly launched processes, but it's incompatible with the PID liveness check — if the PID is dead, there's no race.

### 2. Task-picking ignores orphaned analysing tasks

\`Get-NextWorkflowTask\` in ProcessRegistry.psm1 only picks up \`analysing\` tasks that have \`questions_resolved\` (tasks returning from needs-input). An orphaned analysing task without pending questions is invisible to the task picker.

## Affected Files

- \`workflows/default/systems/runtime/modules/task-reset.ps1\` — lines 252, 267-270 (staleness buffer)
- \`workflows/default/systems/runtime/modules/ProcessRegistry.psm1\` — \`Get-NextWorkflowTask\` (analysing task pickup logic)

### Fix 1: Trust the PID liveness check (task-reset.ps1)

If the process is confirmed dead (PID not found), recover the task immediately regardless of \`updated_at\`:

\`\`\`powershell
# After line 264: task is NOT owned by a live process
# If no live process owns it, the process is dead — recover immediately
# The 5-minute buffer is only needed when we can't verify process liveness
\`\`\`

Either remove the staleness check entirely (the PID check is sufficient) or reduce it to 30 seconds as a minimal race guard.

### Fix 2: Reset orphaned analysing tasks to todo (task-reset.ps1)

When a task is recovered, move it back to \`todo/\` (not leave it in \`analysing/\`). This is already the behavior in the existing recovery code at lines 285-331 — it just never reaches that code due to the staleness skip.

## Acceptance Criteria

- [ ] Tasks stuck in \`analysing\` from a dead process are recovered within seconds on restart
- [ ] Recovered tasks move back to \`todo/\` and are picked up by the next run
- [ ] Tasks owned by live processes are NOT prematurely recovered
- [ ] A task killed 10 seconds ago is recovered on the next startup
- [ ] Layers 1-3 tests pass

## Severity

**High** — interrupted workflows cannot be resumed, requiring manual task file moves.

kabaogluemre:
# fix: Resume killed analysing tasks on restart (#214)

Closes #214

## Problem

When a workflow is killed mid-analysis (task in `analysing/` status), restarting the runner does not pick up the interrupted task. It stays stuck in `analysing/` indefinitely, forcing operators to manually move task files between directories.

## Root Cause

`Reset-AnalysingTasks` already performs an authoritative PID-liveness check — if the owning process is confirmed dead, the task can be safely recovered. However, an **additional 5-minute staleness buffer** on `updated_at` overrode this check and skipped tasks that were updated recently (i.e. freshly killed tasks), even when their owning process was proven dead.

The 5-minute buffer was designed as a race guard against freshly launched processes that haven't written their process file yet — but that race window is measured in milliseconds in practice, not minutes.

## Fix

Reduce the staleness buffer from **5 minutes** to **30 seconds**:

- 30 seconds is more than enough to cover the process-registration race
- Killed processes are now recovered within seconds of restart instead of waiting out a 5-minute window
- PID check remains the primary authority; staleness is only a belt-and-suspenders race guard

```diff
- $stalenessThreshold = $now.AddMinutes(-5)
+ $stalenessThreshold = $now.AddSeconds(-30)


andresharpe (approved):
## dotbot-code-review Summary

Reviewed 1 file (`workflows/default/systems/runtime/modules/task-reset.ps1`, +44/-9). 0 findings survived confidence threshold (>=80).

### Prior Agent Review Summary

| Reviewer | Finding | Status | dotbot-code-review verdict |
|----------|---------|--------|---------------------------|
| @copilot-pull-request-reviewer | Description claims "recovered immediately" but staleness buffer still applied | resolved (outdated) | Fixed -- author added `$deadOwnerTaskIds` set + `$ownerConfirmedDead` bypass so dead-owner signal skips the buffer |
| @copilot-pull-request-reviewer | `Reset-AnalysingTasks` only runs at startup; staleness buffer can still strand tasks | resolved (outdated) | Fixed -- same bypass addresses the stated concern exactly |

Unresolved prior comments: 0 | Confirmed: 0 | Disputed: 0 | Already resolved: 2

### Sanity checks performed

- Dead-owner detection scope: pre-existing `if ($proc.status -notin @('running','starting')) { continue }` at line 229 guarantees `$deadOwnerTaskIds` only captures processes that were running/starting but whose PIDs are gone -- i.e. genuine crash/kill signals, not completed processes.
- Tiered recovery order in the main loop is correct: `liveTaskIds.Contains` (skip) -> `deadOwnerTaskIds.Contains` (recover) -> staleness buffer (recover if older than 30s). Mutually exclusive via short-circuit.
- Staleness buffer narrowed from 5 minutes to 30 seconds: still provides the documented race guard for freshly-launched-but-unregistered processes; kill-recovery latency now bounded at ~30s (minus the bypass for dead-owner signal, which is immediate).
- Call sites verified: `Invoke-AnalysisProcess.ps1:42`, `Invoke-ExecutionProcess.ps1:59`, `Invoke-WorkflowProcess.ps1:109` all invoke `Reset-AnalysingTasks` at startup -- the bypass ensures stuck tasks are now recovered on the first restart after a kill, rather than waiting out the old 5-minute buffer.
- CI green on Windows / Ubuntu / macOS.

### Notes

- No automated regression test added. The test pyramid (`tests/Test-*.ps1`) currently has no direct coverage for `Reset-AnalysingTasks`; the task lifecycle tests exercise the MCP `task_mark_analysing` transition but not the recovery-on-restart path. Recommend a follow-up to add a unit test: plant an analysing-status task + a process registry entry with `status=running, pid=<dead-pid>`, invoke `Reset-AnalysingTasks`, assert the task moved to `todo/`. Tracking this as a follow-up rather than blocking this small fix.

Verdict: **ship**

andresharpe (approved):

PR #311, opened 2026-04-23
fix: workflow-generated tasks were not linked to the workflow that created them 

### LINKED ISSUE #301: Tasks generated during a workflow run (kickstart-from-scratch) are not tagged with the workflow name
## Problem

When a workflow runs and its phases create new tasks through `mcp__dotbot__task_create_bulk` (for example the `expand-task-groups.ps1` phase of `kickstart-from-scratch`), the generated task JSON files are written without a `workflow` field (I'm speaking of `subtasks` created by a workflow phase NOT the workflow phase task itself). The task-runner that executes the workflow filters its queue by workflow name, so it cannot see these new tasks. Once the original manifest tasks finish, the runner considers the workflow complete and exits, leaving the generated tasks sitting in `.bot/workspace/tasks/todo/` with nothing to pick them up.

The tasks are orphaned. The user has no UI path to run them, and re-running the workflow only produces more orphans.

Also tasks created from the UI using the available create task button don't have an associated workflow id.

## Relevant code

- `workflows/default/systems/mcp/tools/task-create-bulk/script.ps1`, lines 148 to 170. The created task object has no `workflow` field.
- `workflows/default/systems/mcp/modules/TaskIndexCache.psm1`, lines 740 and 763. The task-runner filter uses strict equality on the `workflow` field.
- `workflows/default/systems/runtime/modules/ProcessTypes/Invoke-WorkflowProcess.ps1`, lines 215 and 242. The runner exits as soon as `Test-WorkflowComplete` returns true for its filter.
- Commit `e2f3b90` (PR #274) removed the generic "Execute Tasks" button, which was the previous manual escape hatch for running untagged tasks.

### What you expected

Tasks generated during a workflow run should be executable without manual intervention, either because they inherit an appropriate workflow tag.

### Steps to reproduce

1. Run `kickstart-from-scratch` from the UI with a minimal prompt.
2. Wait for the four phases to complete. The workflow tile flips to `completed`.
3. Inspect `.bot/workspace/tasks/todo/*.json`. The files created by phase 4 have no `workflow` property.
4. Click Run on any installed workflow. None of them see the orphan tasks, because each workflow runner filters strictly by its own name.

### Environment

OS: Windows 11

### Severity

high

### Logs / screenshots

All tasks appear when you filter using 'ALL WORKFLOWS'

Image

When using the filter for a workflow only the tasks inside the workflow file will appear (not the created sub-tasks by the workflow phase)

Image

gmireles-ap:
## Linked issue

Closes #301

## Summary of changes

Tasks generated by a running workflow were being written without a `workflow` field, so there was no way to attribute them back to the workflow that produced them.

**Root cause:** the workflow name is a runtime property of the process tree (set via `launch-process.ps1 -Workflow <name>`), but the `task_create` / `task_create_bulk` MCP tool schemas don't declare a `workflow` input at all. There was no mechanism for a caller invoking those tools from inside a workflow run to attach the current workflow name — the information simply wasn't being propagated from the runtime into the MCP call path.

The fix threads the workflow name from the runtime through to the dispatcher via an environment variable:

- **`launch-process.ps1`** now exports `DOTBOT_CURRENT_WORKFLOW` into the process tree, set from the `-Workflow` parameter. The assignment is unconditional (empty string when no workflow is active) so that a value inherited from an outer shell can't contaminate a subsequent run.
- **`dotbot-mcp.ps1`** (tool dispatcher) reads `DOTBOT_CURRENT_WORKFLOW` and merges it into the tool arguments just before dispatch: top-level `workflow` for `task_create`, and `workflow` on each element of `tasks` for `task_create_bulk`. Only fills the field when the caller hasn't supplied one, so explicit callers (e.g. scripts that set `workflow` deliberately) are unaffected.

The `task-create` script already accepts a `workflow` field via its passthrough logic for extra arguments, so no schema or tool-script changes were needed — the problem was purely that the value never reached the dispatcher.

## Screenshots / recordings

N/A — runtime/dispatcher change, no UI surface.

## Testing notes

1. Start a workflow run via `launch-process.ps1 -Workflow <name>` that generates tasks (e.g. a `task_gen` step calling `task_create_bulk`).
2. Inspect the resulting task JSON files in `.bot/workspace/tasks/todo/` and confirm each has `"workflow": "<name>"`.
3. Run a caller that passes `workflow` explicitly and confirm it wins over the env var.
4. Start a shell with `DOTBOT_CURRENT_WORKFLOW` pre-set, then launch a process without `-Workflow`; confirm the child process sees an empty value and generated tasks aren't mis-attributed to the stale name.

## Checklist

- [x] Tests added or updated
- [x] Linked issue exists


gmireles-ap:
Close this PR as #325 closes #301

PR #355, opened 2026-04-27
Bugfix framework integrity paths

### LINKED ISSUE #354: `dotbot init --force` silently fails to commit framework updates due to stale entries in $script:ProtectedPaths
### What happened

`$script:ProtectedPaths` in `core/mcp/modules/FrameworkIntegrity.psm1` still lists `.bot/recipes` and `.bot/workflow.yaml`, both retired by #345 and #347. Neither path is created in current installs.

`init --force` (`scripts/init-project.ps1:1266-1271`) runs `git add -- @stagePaths`, which aborts on the first missing pathspec (`fatal: pathspec '.bot/recipes' did not match any files`) and stages nothing. The follow-up `git commit` then exits non-zero with `no changes added to commit`. `Submit-ForceCommit` suppresses git's stderr via `2>$null`, so the only signal is:

```
Committing framework file updates...
⚠ Framework update commit failed — run manually: DOTBOT_FORCE_COMMIT=1 git commit
```

The framework changes stay uncommitted, which then trips the integrity gate on the next workflow run and surfaces as a misleading `Task '<id>' not found in statuses: analysing, needs-input, analysed`.

### What you expected

`init --force` should commit framework updates cleanly with `chore: update dotbot framework files`. A single missing protected path shouldn't abort the entire stage operation, and stderr suppression shouldn't mask the actual git error.

### Steps to reproduce

1. Fresh project, any current dotbot install — `dotbot init` from a new repo.
2. From the target project: `dotbot init --force`.
3. See `⚠ Framework update commit failed`. The manual `DOTBOT_FORCE_COMMIT=1 git commit` reveals `fatal: pathspec '.bot/recipes' did not match any files`.

### Environment

- OS: Windows 11 (PowerShell 7+, Git Bash for hooks)
- dotbot: v3.5.1, post-#347 `core/` layout
- Reproduces on any fresh project regardless of init history

### Severity

medium

### Logs / screenshots

_No response_

gmireles-ap:
## Linked issue

Closes #354

## Summary of changes

`$script:ProtectedPaths` still listed `.bot/workflow.yaml`, both retired by #345 and #347. On `dotbot init --force`, `git add -- @stagePaths` aborted on the first missing pathspec (`fatal: pathspec '.bot/workflow.yaml' did not match any files`), leaving nothing staged. The follow-up commit then exited non-zero with "no changes added to commit", and `Submit-ForceCommit`'s `2>$null` swallowed the real error — surfacing only the generic `⚠ Framework update commit failed` warning. The uncommitted framework state then tripped the integrity gate on the next workflow run.

- `core/mcp/modules/FrameworkIntegrity.psm1` — drop `.bot/workflow.yaml` from `$script:ProtectedPaths` (no longer created post-refactor).
- `scripts/init-project.ps1` — defensively filter `$stagePaths` to paths that exist on disk, so a single stale entry can't silently abort the stage operation again.

## Screenshots / recordings

N/A — CLI fix, no UI surface.

## Testing notes

1. Fresh target project, any current dotbot install:
   ```pwsh
   mkdir test-project; cd test-project; git init
   dotbot init
   ```
2. From the target project: `dotbot init --force`.
3. **Before this PR:** `⚠ Framework update commit failed`; running the suggested `DOTBOT_FORCE_COMMIT=1 git commit` reveals `fatal: pathspec '.bot/workflow.yaml' did not match any files`.
4. **After this PR:** `✓ Framework update committed`. `git log -1` shows the `chore: update dotbot framework files` commit, `git status .bot/` is clean.

## Checklist

- [x] Linked issue exists

### DIFF
--- a/core/mcp/modules/FrameworkIntegrity.psm1
+++ b/core/mcp/modules/FrameworkIntegrity.psm1
@@ -23,6 +23,7 @@
 $script:ProtectedPaths = @(
     '.bot/core',
     '.bot/hooks',
+    '.bot/recipes',
     '.bot/settings/providers',
     '.bot/settings/settings.default.json',
     '.bot/settings/theme.default.json',
--- a/core/mcp/modules/FrameworkIntegrity.psm1
+++ b/core/mcp/modules/FrameworkIntegrity.psm1
@@ -23,13 +23,11 @@
 $script:ProtectedPaths = @(
     '.bot/core',
     '.bot/hooks',
-    '.bot/recipes',
     '.bot/settings/providers',
     '.bot/settings/settings.default.json',
     '.bot/settings/theme.default.json',
     '.bot/go.ps1',
     '.bot/init.ps1',
-    '.bot/workflow.yaml',
     '.bot/.gitignore',
     '.bot/.manifest.json',
     '.bot/README.md'
--- a/scripts/init-project.ps1
+++ b/scripts/init-project.ps1
@@ -1263,7 +1263,15 @@ if ($LASTEXITCODE -ne 0) {
     # then commit framework paths + manifest if anything actually changed.
     New-FrameworkManifest -Root $ProjectDir -Generator 'dotbot init --force' -Paths $frameworkPaths
 
-    $stagePaths = $frameworkPaths + @('.bot/.manifest.json')
+    # Filter to paths that actually exist on disk — git add chokes on a single
+    # missing pathspec (`fatal: pathspec '...' did not match any files`) and
+    # aborts the entire stage operation, which then surfaces as a silent
+    # "Framework update commit failed" because `Submit-ForceCommit` runs git
+    # commit with nothing staged. Filtering here keeps a single stale entry
+    # in $script:ProtectedPaths from breaking init --force.
+    $stagePaths = @(($frameworkPaths + @('.bot/.manifest.json')) | Where-Object {
+        Test-Path -LiteralPath (Join-Path $ProjectDir $_)
+    })
     $dirty = git -C $ProjectDir status --porcelain -- @stagePaths 2>$null
     if ($dirty) {
         Write-DotbotCommand "Committing framework file updates..."
--- a/scripts/init-project.ps1
+++ b/scripts/init-project.ps1
@@ -1263,14 +1263,14 @@ if ($LASTEXITCODE -ne 0) {
     # then commit framework paths + manifest if anything actually changed.
     New-FrameworkManifest -Root $ProjectDir -Generator 'dotbot init --force' -Paths $frameworkPaths
 
-    # Filter to paths that actually exist on disk — git add chokes on a single
-    # missing pathspec (`fatal: pathspec '...' did not match any files`) and
-    # aborts the entire stage operation, which then surfaces as a silent
-    # "Framework update commit failed" because `Submit-ForceCommit` runs git
-    # commit with nothing staged. Filtering here keeps a single stale entry
-    # in $script:ProtectedPaths from breaking init --force.
+    # Keep paths that are on disk OR tracked in git. Drops stale list entries
+    # (would abort `git add` with a pathspec error), but preserves
+    # tracked-then-deleted paths so migration deletions still get staged.
     $stagePaths = @(($frameworkPaths + @('.bot/.manifest.json')) | Where-Object {
-        Test-Path -LiteralPath (Join-Path $ProjectDir $_)
+        $abs = Join-Path $ProjectDir $_
+        if (Test-Path -LiteralPath $abs) { return $true }
+        $tracked = git -C $ProjectDir ls-files -- $_ 2>$null
+        return [bool]$tracked
     })
     $dirty = git -C $ProjectDir status --porcelain -- @stagePaths 2>$null
     if ($dirty) {
--- a/tests/Test-Components.ps1
+++ b/tests/Test-Components.ps1
@@ -4427,10 +4427,9 @@ if ((Test-Path $manifestModule) -and (Test-Path $frameworkIntegrityModule)) {
         & git config user.email "test@test.com" 2>$null
         & git config user.name "Test" 2>$null
 
-        # Create a .bot/ structure with two protected dirs and a protected file.
-        # Include the sentinel file (dotbot-mcp.ps1) at .bot/core/mcp/ that
-        # Test-FrameworkIntegrity uses to detect pre-first-commit state via git log.
-        $protectedPaths = @('.bot/core', '.bot/go.ps1')
+        # Fixture: dotbot-mcp.ps1 is the sentinel Test-FrameworkIntegrity probes
+        # for pre-first-commit detection; .bot/go.ps1 is the tampering target.
+        $protectedPaths = Get-FrameworkProtectedPaths
         New-Item -ItemType Directory -Path (Join-Path $fiTestDir ".bot/core/mcp") -Force | Out-Null
         Set-Content -Path (Join-Path $fiTestDir ".bot/core/mcp/dotbot-mcp.ps1") -Value "# mcp server" -Encoding UTF8
         Set-Content -Path (Join-Path $fiTestDir ".bot/go.ps1") -Value "# go" -Encoding UTF8
--- a/tests/Test-Structure.ps1
+++ b/tests/Test-Structure.ps1
@@ -369,16 +369,23 @@ if (-not $dotbotInstalled) {
         Assert-PathExists -Name "-Force: .control/settings.json preserved" -Path $dummySettings
         Assert-PathExists -Name "-Force: system files refreshed" -Path (Join-Path $botDir "core/mcp/dotbot-mcp.ps1")
 
-        # Regression guard: init --force must leave a clean .bot/ tree, else
-        # the next workflow run's integrity gate trips with "tampered".
-        Push-Location $testProject
-        try {
-            $dirtyAfterForce = & git status --porcelain -- .bot/ 2>$null
-            Assert-True -Name "-Force: clean working tree (no uncommitted .bot/ changes)" `
-                -Condition ([string]::IsNullOrWhiteSpace(($dirtyAfterForce -join "`n"))) `
-                -Message "init --force left uncommitted .bot/ changes:`n$($dirtyAfterForce -join "`n")"
-        } finally {
-            Pop-Location
+        # Regression guard: init --force must leave a clean framework tree,
+        # else the next workflow run's integrity gate trips with "tampered".
+        # Scope to the protected-paths list — workspace/ and .control/ hold
+        # user/runtime data the test deliberately seeds and aren't framework.
+        $integrityModule = Join-Path $dotbotDir "core/mcp/modules/FrameworkIntegrity.psm1"
+        if (Test-Path $integrityModule) {
+            Import-Module $integrityModule -Force
+            $protectedPaths = @(Get-FrameworkProtectedPaths)
+            Push-Location $testProject
+            try {
+                $dirtyFramework = & git status --porcelain -- @protectedPaths 2>$null
+                Assert-True -Name "-Force: clean framework tree (no uncommitted protected-path changes)" `
+                    -Condition ([string]::IsNullOrWhiteSpace(($dirtyFramework -join "`n"))) `
+                    -Message "init --force left uncommitted framework changes:`n$($dirtyFramework -join "`n")"
+            } finally {
+                Pop-Location
+            }
         }
 
         if ($initialInstanceId) {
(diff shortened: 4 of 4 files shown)


andresharpe (changes requested):
Thanks for the fix - the underlying bug (silent commit failure leaving the tree dirty) is real and worth fixing. Two things to revise based on the Copilot review, plus one tweak to the description.

## 1. `.bot/recipes` is not retired - keep it in `ProtectedPaths`

Copilot is right here. Evidence:

- Three stacks populate `.bot/recipes/` on init: `stacks/dotnet/recipes/skills/` (eleven SKILL.md files), `stacks/dotnet-blazor/recipes/skills/`, `stacks/dotnet-ef/recipes/skills/`.
- `core/init.ps1:54-55` references `.bot/recipes/agents` and `.bot/recipes/skills` for IDE setup.
- `scripts/init-project.ps1:1303` still announces `.bot/recipes/` to users at end of init.
- `scripts/init-project.ps1:747` has special-case logic that lets stack recipes through to `.bot/recipes`.

I suggest you restore `.bot/recipes` to `$script:ProtectedPaths` and rely on the staging-side filter to tolerate its absence on bare projects.

## 2. `.bot/workflow.yaml` is genuinely retired - removal there is correct

Counter-evidence so you can keep it removed with confidence:

- `scripts/init-project.ps1:218-225` actively deletes `.bot/workflow.yaml` in the migration step.
- `scripts/init-project.ps1:269` includes it in the legacy root files removed under `--force`.
- `tests/Test-WorkflowIntegration.ps1:78` asserts "Default init: no .bot/workflow.yaml at bot root".

So the PR description's claim covers this path correctly, just not the recipes one.

## 3. Filter needs to keep tracked-then-deleted paths

The `Test-Path` filter has the gap Copilot called out. If `--force` rewrites a framework path away (file deleted from disk, still in git index), `Test-Path` returns false, the path is filtered out, and `git status --porcelain -- @stagePaths` never sees the deletion. The result is no commit, dirty tree, integrity gate trips next run - the same failure mode this PR set out to fix.

Suggested shape: keep paths that exist on disk OR are tracked in git.

```powershell
$stagePaths = @(($frameworkPaths + @('.bot/.manifest.json')) | Where-Object {
    $abs = Join-Path $ProjectDir $_
    if (Test-Path -LiteralPath $abs) { return $true }
    $tracked = git -C $ProjectDir ls-files -- $_ 2>$null
    return [bool]$tracked
})
```

I would avoid `git ls-files --error-unmatch` here - it writes to stderr and exits non-zero on a miss, forcing extra discipline. Plain `ls-files -- <path>` returns empty + exit 0 for an unknown path and lines for tracked-or-tracked-then-deleted paths, for both file and directory pathspecs. Check the output, not `$LASTEXITCODE`.

## 4. PR description tweak

The description says both paths were "retired by #345/#347". Only `.bot/workflow.yaml` was. I suggest narrowing the description to that one and reframing the fix as: filter staging paths to the union of present-on-disk and tracked-in-git, so a single missing optional path can no longer abort the stage operation.

Optional follow-up worth filing separately (not blocking this PR): `Submit-ForceCommit` swallows real errors via `2>$null`. The original investigation only surfaced the root cause because someone ran the suggested manual `git commit` - a less hardy contributor would be stuck. Worth surfacing the underlying git error in the warning.

core/mcp/modules/FrameworkIntegrity.psm1
```diff
@@ -23,13 +23,11 @@
 $script:ProtectedPaths = @(
     '.bot/core',
     '.bot/hooks',
```
andresharpe: [BLOCKER] Confirming @copilot's finding here with additional evidence on why this matters.

Three independent defense layers all draw from $frameworkPaths (= Get-FrameworkProtectedPaths):

1. **Pre-commit hook case patterns** -- scripts/init-project.ps1:1077-1090 regenerates the hook's shell case statement from $frameworkPaths on every dotbot init. With .bot/recipes removed here, the regenerated hook no longer contains a 
ecipes/* arm, so direct commits that modify .bot/recipes/skills/**/SKILL.md (or any stack-installed recipe) bypass the framework-file protection at line 1136 and require no DOTBOT_FORCE_COMMIT=1 bypass.
2. **SHA256 manifest entries** -- New-FrameworkManifest writes hashes only for paths in $frameworkPaths. Recipe files become unhashed, so committed tampering (e.g. via --no-verify) is invisible to the manifest layer.
3. **Integrity gate scan** -- Test-FrameworkIntegrity git-status check no longer covers .bot/recipes.

Restoring .bot/recipes to $script:ProtectedPaths (as @copilot suggested) fixes all three layers at once. The staging-side filter then handles the no-stack case where the directory does not exist on disk.

`suggestion
    '.bot/hooks',
    '.bot/recipes',
`
"@
        },
        @{
            path = "scripts/init-project.ps1"
            line = 1272
            side = "RIGHT"
            body = @"
[MAJOR] Missing regression tests for the bug this PR fixes.

The PR fixes a regression where dotbot init --force silently leaves the working tree dirty when $script:ProtectedPaths contains a stale entry. Two existing tests should be extended so the next drift is caught automatically:

1. **	ests/Test-Structure.ps1:364-377** -- the -Force block currently only asserts file presence and instance_id preservation. It never runs git status --porcelain to verify the framework commit actually completed. A clean working tree after init --force assertion would have caught this bug directly.

2. **	ests/Test-Components.ps1:4433** -- the FrameworkIntegrity behavioral tests use synthetic protectedPaths = @('.bot/core', '.bot/go.ps1'). They never exercise the real Get-FrameworkProtectedPaths list, so a stale entry there is invisible to the test layer. A Layer 2 assertion that every entry in the real list resolves to a path under a freshly-initialised .bot/ would prevent recurrence.

Both tests live outside this diff, so I cannot attach a suggestion block here. Sketch for Test-Structure.ps1 to anchor the discussion:

`
Push-Location 
 = & git status --porcelain 2>
Pop-Location
Assert-True -Name "-Force: working tree clean after re-init" `
    -Condition ([string]::IsNullOrWhiteSpace(($porcelain -join "`n"))) `
    -Message "Expected clean tree, got:`n$($porcelain -join "`n")"
`

Per CLAUDE.md test pyramid, this belongs in Layer 1 (Test-Structure) plus Layer 2 (Test-Components).

andresharpe (commented):
## dotbot-code-review supplement

Reviewed two changed files. Two novel findings beyond the prior @copilot[bot] threads and my earlier request-changes review.

### Prior agent review summary

| Agent | File | Finding | Agrees? |
|-------|------|---------|---------|
| @copilot[bot] | FrameworkIntegrity.psm1:L25 | .bot/recipes removal weakens integrity | Yes -- confirming with new evidence (pre-commit hook also regenerates) |
| @copilot[bot] | init-project.ps1:L1274 | Test-Path filter drops deletions | Yes -- already covered in my prior review |
| @andresharpe (top-level) | n/a | restore .bot/recipes; ls-files filter; description fix | Already posted |

Unresolved prior comments reviewed: 2 | Confirmed: 2 | Disputed: 0

Verdict: fix majors first

scripts/init-project.ps1
```diff
@@ -1263,7 +1263,15 @@ if ($LASTEXITCODE -ne 0) {
...
     New-FrameworkManifest -Root $ProjectDir -Generator 'dotbot init --force' -Paths $frameworkPaths
 
-    $stagePaths = $frameworkPaths + @('.bot/.manifest.json')
+    # Filter to paths that actually exist on disk — git add chokes on a single
+    # missing pathspec (`fatal: pathspec '...' did not match any files`) and
+    # aborts the entire stage operation, which then surfaces as a silent
+    # "Framework update commit failed" because `Submit-ForceCommit` runs git
+    # commit with nothing staged. Filtering here keeps a single stale entry
+    # in $script:ProtectedPaths from breaking init --force.
+    $stagePaths = @(($frameworkPaths + @('.bot/.manifest.json')) | Where-Object {
```
andresharpe, line 1269: [MAJOR] Missing regression tests for the bug this PR fixes.

The PR fixes a regression where `dotbot init --force` silently leaves the working tree dirty when `$script:ProtectedPaths` contains a stale entry. Two existing tests should be extended so the next drift is caught automatically:

1. **`tests/Test-Structure.ps1:364-377`** -- the `-Force` block currently only asserts file presence and `instance_id` preservation. It never runs `git status --porcelain` to verify the framework commit actually completed. A "clean working tree after init --force" assertion in this block would have caught this bug directly.

2. **`tests/Test-Components.ps1:4433`** -- the `FrameworkIntegrity` behavioral tests use a synthetic `protectedPaths = @('.bot/core', '.bot/go.ps1')`. They never exercise the real `Get-FrameworkProtectedPaths` list, so stale entries there are invisible to the test layer. A Layer 2 assertion that every entry in the real list resolves to a path under a freshly-initialised `.bot/` would prevent recurrence.

Both tests live outside this diff so I cannot attach a `suggestion` block here. Per the CLAUDE.md test pyramid, the new assertions belong in Layer 1 (Test-Structure) plus Layer 2 (Test-Components).

core/mcp/modules/FrameworkIntegrity.psm1
andresharpe, line 25: [BLOCKER] Confirming @copilot's finding here with additional evidence on why this matters.

Three independent defense layers all draw from `$frameworkPaths` (= `Get-FrameworkProtectedPaths`):

1. **Pre-commit hook case patterns** -- `scripts/init-project.ps1:1077-1090` regenerates the hook's shell case statement from `$frameworkPaths` on every `dotbot init`. With `.bot/recipes` removed here, the regenerated hook no longer contains a `recipes/*` arm, so direct commits that modify `.bot/recipes/skills/**/SKILL.md` (or any stack-installed recipe) bypass the framework-file protection at line 1136 and require no `DOTBOT_FORCE_COMMIT=1` bypass.
2. **SHA256 manifest entries** -- `New-FrameworkManifest` writes hashes only for paths in `$frameworkPaths`. Recipe files become unhashed, so committed tampering (e.g. via `--no-verify`) is invisible to the manifest layer.
3. **Integrity gate scan** -- `Test-FrameworkIntegrity` git-status check no longer covers `.bot/recipes`.

Restoring `.bot/recipes` to `$script:ProtectedPaths` (as @copilot suggested) fixes all three layers at once. The staging-side filter then handles the no-stack case where the directory does not exist on disk.

```suggestion
    '.bot/hooks',
    '.bot/recipes',
```

gmireles-ap:
Addressed the review.

The .bot/recipes path stays protected. The dotnet stacks populate it on install, which I had incorrectly assumed was moved or deleted because there's no recipes folder after a plain `dotbot init`. Thanks for the catch.

The staging filter is more robust now: when --force migrates legacy content away, deletions still get staged.

Both regression assertions you asked for are in place — both at Layer 1 — and the Layer 2 behavioral test has been extended to use the full list (that last change doesn't catch any new bug class; it just makes more sense to exercise the real list). The -Force block now checks that the framework tree is fully committed afterward, which directly catches the silent-commit-failure mode this PR fixes. The --Stack dotnet block now also asserts that every entry in the real Get-FrameworkProtectedPaths resolves on disk in a stacked init, which catches stale entries no install path actually creates. Both placed at Layer 1 because Layer 2's golden snapshots only cover workflow flavors today, not stacks — and only stacks populate `.bot/recipes`. And also because checking that there are not stale paths is more of a 'Structural Change' than a 'Behavioral one'.

Only `.bot/workflow.yaml` is genuinely retired.

core/mcp/modules/FrameworkIntegrity.psm1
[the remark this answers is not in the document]
gmireles-ap: Suggestion applied by me.

gmireles-ap: Agree. I apologize, the `.bot/recipes` indeed has not been deleted. It was only deleted by default, still exists but only when the user adds a stack.

scripts/init-project.ps1
gmireles-ap: 1) Added.

2) Added, but in Test-Structure, and extended tests/Test-Components.ps1:4433 to use the full list — for two reasons: (a) Layer 1 already verifies the install's file structure, so checking that the protected-paths list matches that structure fits the same layer; (b) the existing behavioral test in Layer 2 verifies that protected paths actually get protected (regardless of how many or which ones; I extended it to the full list anyway so it aligns better), whereas checking whether the list itself is correct is a structural concern, not behavioral (I'm not the best person to weigh in on semantics, tbh).

andresharpe: please esolve conversation if you are happy it is handled?

andresharpe (approved):
Final /pwsh-review pass: 0 findings after calibration. Static analysis clean (5 template-substitution findings out of PR scope). All 5 review threads resolved. CI green. New regression tests cover both clean-framework-tree and protected-paths-resolve scenarios.

Note: the diff-bug agent flagged (then calibrator dropped at conf 70) a potential cosmetic side-effect — for users upgrading from a pre-#392 install where .bot/workflow.yaml was tracked, init --force will leave a dangling 'D .bot/workflow.yaml' in git status. Not a regression vs prior behavior, just unsurfaced by the new regression test. Worth a one-line follow-up issue if it shows up in upgrade reports.

PR #385, opened 2026-04-29
fix(runtime): pin claude cwd to task worktree

### LINKED ISSUE #314: Worktree isolation broken: agent edits land on project root instead of task worktree
### What happened

The framework creates a per-task git worktree at `../worktrees/{repo}/task-{id}-{slug}/`, but the Claude CLI child process is always spawned with `WorkingDirectory` set to `$global:DotbotProjectRoot`. The worktree path is never threaded through to the process.

Specifically:
- `Invoke-WorkflowProcess.ps1` calls `Push-Location $worktreePath` before invoking the provider, but `Push-Location` only affects the parent PowerShell shell. `System.Diagnostics.Process` uses its own `ProcessStartInfo.WorkingDirectory`, so the change does not propagate to the child.
- `ClaudeCLI.psm1` hardcodes `$psi.WorkingDirectory = $global:DotbotProjectRoot` and exposes no `-WorkingDirectory` parameter on `Invoke-ClaudeStream`.
- `ProviderCLI.psm1` has no `-WorkingDirectory` parameter on `Invoke-ProviderStream` and no passthrough either.

As a result, every `Edit` / `Write` / `Bash` call the agent makes during execution operates on the project root (usually on `main`), not on the task branch inside the worktree. The worktree stays empty; the task branch has no real commits. When the task is "completed", the squash-merge carries a file diff that does not exist in the task branch's own history — it leaks in through the stash/pop cycle around the merge.

### What you expected

During the execution phase, the Claude CLI process should be spawned with its working directory set to the task's worktree path, so that every tool call (`Edit`, `Write`, `Bash`, etc.) operates on the task branch in isolation. The analysis phase can keep running from project root because the worktree does not exist yet at that point. On completion, the squash-merge into `main` should reflect commits that genuinely live on the task branch.

### Steps to reproduce

1. In a clean test repo, run `dotbot init`. Make sure a tracked `README.md` exists.
2. Create a task:
   ```bash
   dotbot task create --name "Append line" --description "Add the line 'hello from dotbot' at the end of README.md"
   ```
3. Run the task autonomously.
4. Before the task finishes, compare the two locations:
   ```bash
   grep "hello from dotbot" ../worktrees/{repo}/task-*/README.md   # expected: match, actual: no match
   grep "hello from dotbot" ./README.md                            # expected: no match, actual: match
   ```
5. After the task finishes, inspect the squash-merge commit on `main`: the file diff is present, but the corresponding change is not in the task branch's commit history — it was picked up through the stash/pop around the merge, not produced by the agent on the task branch.

### Environment

All environments

### Severity

medium

### Logs / screenshots

_No response_

kabaogluemre:
## Summary

Closes #314.

The Claude child process was always spawned with `WorkingDirectory = $global:DotbotProjectRoot`, so during the execution phase agent `Edit` / `Write` / `Bash` calls landed on the project root instead of the per-task git worktree. The wrong-folder edits then leaked into the squash-merge via the stash/pop cycle around merge, leaving the task branch with no real history of its own.

## Why `Push-Location` alone wasn't enough

`Push-Location $worktreePath` only updates PowerShell's `$PWD`. `System.Diagnostics.Process` reads `ProcessStartInfo.WorkingDirectory` directly and ignores the shell's location stack, so the Claude process inherited the server's cwd regardless of the parent shell.

PowerShell's `&` call operator *does* honour `$PWD`, which is why the existing `Push-Location` is intentionally kept — it covers non-Claude providers (codex/gemini) that spawn via `& $executable`.

## What changed

| File | Change |
|---|---|
| `core/runtime/ClaudeCLI/ClaudeCLI.psm1` | New `[string]$WorkingDirectory` parameter on `Invoke-ClaudeStream`. When set, it overrides `psi.WorkingDirectory`; otherwise the existing `$global:DotbotProjectRoot` fallback is used. |
| `core/runtime/ProviderCLI/ProviderCLI.psm1` | `Invoke-ProviderStream` accepts `[string]$WorkingDirectory` and forwards it into `$streamArgs` for the Claude branch. |
| `core/runtime/modules/ProcessTypes/Invoke-WorkflowProcess.ps1` | Execution-phase call site passes `WorkingDirectory = $worktreePath`. Analysis-phase call site adds a comment explaining why it does **not** pass one (worktree is created after analysis; analysis is read-only by prompt contract). |
| `tests/mock-claude.ps1` | One-line addition: writes `(Get-Location).Path` to `$DOTBOT_MOCK_LOG_DIR/mock-claude-cwd.log` so tests can assert the spawned cwd. |
| `tests/Test-MockClaude.ps1` | New Layer 3 section "WORKING DIRECTORY (#314)" with three regression assertions. |

## Scope notes

- **Analysis phase is intentionally untouched.** The worktree is created at `New-TaskWorktree` (line ~1421), after the analysis call site at line 1256. Analysis stays at project root, which is correct for its read-only contract.
- **`.mcp.json` discovery still works** in the worktree: it's gitignored at the project root and `Copy-BuildArtifacts` (`WorktreeManager.psm1:992`) copies it into the worktree at creation time alongside other gitignored top-level files.
- **The stash/pop / squash-merge interaction is out of scope** — per @carlospedreira's comment, that's a separate question that needs its own focused reproduction. This PR only fixes the confirmed root cause: Claude editing the wrong folder.

## Test plan

New Layer 3 assertions in `tests/Test-MockClaude.ps1` (under "WORKING DIRECTORY (#314)"):

- [x] `Invoke-ClaudeStream -WorkingDirectory <dir>` -> child cwd equals `<dir>`
- [x] `Invoke-ClaudeStream` without `-WorkingDirectory` -> child cwd falls back to `$global:DotbotProjectRoot`
- [x] `Invoke-ProviderStream -WorkingDirectory <dir>` (Claude branch) -> forwarded to `Invoke-ClaudeStream`, child cwd equals `<dir>`

### DIFF
--- a/tests/Test-MockClaude.ps1
+++ b/tests/Test-MockClaude.ps1
@@ -202,14 +202,29 @@ try {
         $tempCwd = Join-Path ([System.IO.Path]::GetTempPath()) "dotbot-cwd-$([System.Guid]::NewGuid().ToString('N').Substring(0,8))"
         New-Item -Path $tempCwd -ItemType Directory -Force | Out-Null
 
-        # Resolve to canonical form (Windows temp paths can include short-name segments
-        # like RUNNER~1 vs runneradmin; the mock captures (Get-Location).Path which is
-        # already long-form, so we compare against the long-form too).
-        $expectedCwd = (Resolve-Path -LiteralPath $tempCwd).Path
+        # Canonicalize to match what the kernel reports as cwd inside the spawned process.
+        # - Windows: Resolve-Path expands short-name segments (RUNNER~1 -> runneradmin).
+        # - macOS:   /var, /tmp, /etc are symlinks to /private/*. getcwd() in the child
+        #            returns the resolved /private/* form, so we follow links here too.
+        # - Linux:   pwd -P resolves any symlinks in path components.
+        # Resolve-Path alone does not follow symlinks, so on POSIX we shell out to pwd -P.
+        function Get-CanonicalCwd {
+            param([Parameter(Mandatory)][string]$Path)
+            $resolved = (Resolve-Path -LiteralPath $Path).Path
+            if ($IsMacOS -or $IsLinux) {
+                $shellResolved = & /bin/sh -c "cd `"$resolved`" && pwd -P" 2>$null
+                if ($LASTEXITCODE -eq 0 -and $shellResolved) {
+                    $resolved = $shellResolved.Trim()
+                }
+            }
+            return $resolved
+        }
+
+        $expectedCwd = Get-CanonicalCwd -Path $tempCwd
 
         # Save and rebuild $global:DotbotProjectRoot so the fallback assertion is deterministic
         $savedDotbotProjectRoot = $global:DotbotProjectRoot
-        $global:DotbotProjectRoot = (Resolve-Path -LiteralPath (Split-Path -Parent $dotbotDir)).Path
+        $global:DotbotProjectRoot = Get-CanonicalCwd -Path (Split-Path -Parent $dotbotDir)
 
         try {
             # 1. -WorkingDirectory pins the child cwd
--- a/tests/Test-MockClaude.ps1
+++ b/tests/Test-MockClaude.ps1
@@ -191,6 +191,78 @@ try {
     Write-Host ""
 
     # ═══════════════════════════════════════════════════════════════════
+    # WORKING DIRECTORY (#314)
+    # ═══════════════════════════════════════════════════════════════════
+
+    Write-Host "  WORKING DIRECTORY (#314)" -ForegroundColor Cyan
+    Write-Host "  ────────────────────────────────────────────" -ForegroundColor DarkGray
+
+    if (Test-Path $claudeModule) {
+        $cwdLog = Join-Path $mockLogDir "mock-claude-cwd.log"
+        $tempCwd = Join-Path ([System.IO.Path]::GetTempPath()) "dotbot-cwd-$([System.Guid]::NewGuid().ToString('N').Substring(0,8))"
+        New-Item -Path $tempCwd -ItemType Directory -Force | Out-Null
+
+        # Resolve to canonical form (Windows temp paths can include short-name segments
+        # like RUNNER~1 vs runneradmin; the mock captures (Get-Location).Path which is
+        # already long-form, so we compare against the long-form too).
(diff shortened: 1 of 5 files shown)


tests/Test-MockClaude.ps1
```diff
@@ -190,6 +190,93 @@ try {
...
+        $savedDotbotProjectRoot = $global:DotbotProjectRoot
+        $global:DotbotProjectRoot = Get-CanonicalCwd -Path (Split-Path -Parent $dotbotDir)
+
+        try {
+            # 1. -WorkingDirectory pins the child cwd
+            try {
+                Invoke-ClaudeStream -Prompt "cwd test explicit" -Model "opus" -WorkingDirectory $tempCwd *>&1 | Out-Null
+                $captured = if (Test-Path $cwdLog) { (Get-Content $cwdLog -Raw).Trim() } else { "" }
+                Assert-True -Name "Invoke-ClaudeStream pins cwd to -WorkingDirectory (#314)" `
+                    -Condition ($captured -ieq $expectedCwd) `
```
andresharpe: **[nit] (80)** [PWSH-LANG-CASE-INSENSITIVE-PATH-COMPARE]

Case-insensitive path comparison hides POSIX bugs. All three new cwd assertions (lines 235, 246, 260) compare paths via `-ieq`. On Linux and macOS the filesystem is case-sensitive, so `-ieq` would mark `/tmp/Foo` and `/tmp/foo` as matching even if a plumbing bug accidentally lowercased a path component. The CI matrix includes core-7.4-linux and core-7.4-macos -- the platforms where case-mangling is most likely to surface.

Defeats the test's value as a regression guard for case-mangling cwd plumbing on the platforms where case-mangling is most likely to manifest.

Fix: Gate on $IsWindows: use `-ieq` on Windows (case-insensitive filesystem) and `-ceq` on POSIX.

```suggestion
$pathsMatch = if ($IsWindows) { $captured -ieq $expectedCwd } else { $captured -ceq $expectedCwd }
Assert-True -Name "..." -Condition $pathsMatch ...
```

andresharpe (dismissed):
## Code review summary

_pwsh-code-review v0.19.0_

0 blocker, 0 major, 0 minor, 1 nit, 0 question

Verdict: **ship**

---

andresharpe (approved):
Final review pass clean (multi-agent /pwsh-review: 0 findings; static analysis findings out of PR scope). Doc nit and case-aware comparison nit addressed.

PR #386, opened 2026-04-29
fix(init): merge core MCP entries into existing .mcp.json (#315)

### LINKED ISSUE #315: dotbot init skips existing .mcp.json instead of merging core MCP server entries
### What happened

If a project already has a `.mcp.json` before running `dotbot init`, the init script skips the file entirely instead of merging in the required core entries (`dotbot`, `context7`, `playwright`).

In `scripts/init-project.ps1` around line 815:

```powershell
if (Test-Path $mcpJsonPath) {
    Write-DotbotWarning ".mcp.json already exists -- skipping"
} else {
    # ... writes dotbot + context7 + playwright entries
}
```

There is no merge, and `--force` is not considered in this branch. Any existing `.mcp.json` — e.g. one the user maintains for their own MCP servers like `github`, `figma`, etc. — causes the core entries to never be written. The downstream effect is a hard failure: `pwsh .bot/go.ps1` reports `Dotbot MCP server not registered`, because the doctor check at `init-project.ps1:1304-1320` looks for `mcpServers.dotbot` and does not find it.

This is especially confusing because the codebase already has a working `Merge-McpServers` function, used at `init-project.ps1:429` to merge stack-workflow MCP servers from `manifest.mcp_servers`. The core `dotbot` / `context7` / `playwright` entries just don't go through it — they live in a separate branch that only writes when the file is absent.

### What you expected

`dotbot init` should merge into an existing `.mcp.json` rather than skip it:

- Parse the existing file.
- If a core entry (`dotbot`, `context7`, `playwright`) is missing, add it.
- If a core entry already exists, leave it alone unless `--force` is passed; with `--force`, refresh it to the current canonical definition.
- User-added entries (`github`, `figma`, anything else) must be preserved verbatim in all cases.
- If the existing `.mcp.json` is not valid JSON, fail loudly with an actionable error rather than silently skipping or overwriting.

After a successful `dotbot init` on a project that already had a `.mcp.json`, `pwsh .bot/go.ps1` should start cleanly and the doctor check should find the `dotbot` MCP server registered.

### Steps to reproduce

1. In a clean test repo, write a `.mcp.json` by hand before running init:
   ```json
   { "mcpServers": { "myserver": { "command": "echo", "args": ["hi"] } } }
   ```
2. Run `dotbot init --force`.
3. Inspect `.mcp.json`. Observed: only `myserver` is present; no `dotbot`, `context7`, or `playwright` entry has been added. The console shows the `".mcp.json already exists -- skipping"` warning.
4. Run `pwsh .bot/go.ps1`. Observed: the startup check fails with `Dotbot MCP server not registered`.

Expected after the fix:
- `.mcp.json` contains `myserver` (preserved) plus `dotbot`, `context7`, and `playwright` (merged in).
- `pwsh .bot/go.ps1` starts without the "not registered" error.

### Environment

All environments

### Severity

medium

### Logs / screenshots

_No response_

kabaogluemre:
## Summary

Closes #315 — `dotbot init` skipped existing `.mcp.json` instead of merging the framework's core MCP entries (`dotbot`, `context7`, `playwright`).

If a project already had a user-maintained `.mcp.json` (e.g. `github`, `figma` servers), init printed `".mcp.json already exists -- skipping"` and never wrote the core entries. `--force` did not help — same skip path. Downstream, `pwsh .bot/go.ps1` failed with `Dotbot MCP server not registered`.

## What changed

`scripts/init-project.ps1` — replaced the skip-if-exists branch with a merge:

- **Pre-existing `.mcp.json`** is parsed; invalid JSON now `throw`s with an actionable message instead of being silently skipped or overwritten.
- **Core entries** (`dotbot`, `context7`, `playwright`) are added when missing. If already present, they are left alone unless `-Force` is passed; with `-Force`, they are refreshed to the canonical definition.
- **User-added entries** (`github`, `figma`, anything non-core) are preserved verbatim, after the core entries.
- File is re-emitted with `ConvertTo-Json -Depth 10` so nested args/env survive the round-trip.

Inline merge logic was kept rather than rerouting through `Merge-McpServers` in `core/runtime/modules/workflow-manifest.ps1` — that helper has skip-if-exists semantics by design (used by workflow installs) and no `-Force` refresh path. Bolting both onto it would change its contract for every existing caller (`scripts/workflow-add.ps1`, the workflow loop in `init-project.ps1:507`).

## Behaviour matrix

| Scenario | Before | After |
|----------|--------|-------|
| No `.mcp.json` | core entries written | core entries written (unchanged) |
| `.mcp.json` with user entry only | **skipped** — core missing, `go.ps1` fails | core merged in, user entry preserved |
| `.mcp.json` with tampered core entry, no `-Force` | skipped | tampered entry kept (with note: "use `-Force` to refresh") |
| `.mcp.json` with tampered core entry, `-Force` | skipped | refreshed to canonical |
| Invalid JSON | skipped (silent) | `throw` with `.mcp.json` error message |

## Tests

Added Layer 1 regression cases under **INIT MCP MERGE (#315)** in `tests/Test-Structure.ps1`:

1. Pre-existing user entry (`myserver`) is preserved **and** all three core entries are added.
2. `-Force` refreshes a tampered core entry (`dotbot.command = "WRONG"` → `"pwsh"`).
3. Invalid JSON causes init to fail loudly (non-zero exit or `.mcp.json` referenced in output).

## Test plan

- [ ] `pwsh install.ps1` succeeds
- [ ] `pwsh tests/Run-Tests.ps1` — all of layers 1-3 green, including the new `INIT MCP MERGE (#315)` cases
- [ ] Manual repro from the issue: pre-write a `.mcp.json` with `myserver`, run `dotbot init --force`, confirm `myserver` + `dotbot` + `context7` + `playwright` all present
- [ ] `pwsh .bot/go.ps1` starts cleanly afterwards (no "Dotbot MCP server not registered")
- [ ] Invalid JSON case: pre-write garbage, run init, confirm loud failure rather than silent skip

### DIFF
--- a/scripts/init-project.ps1
+++ b/scripts/init-project.ps1
@@ -876,52 +876,97 @@ if (Test-Path $initScript) {
 }
 
 # ---------------------------------------------------------------------------
-# Create .mcp.json with MCP server configuration
+# Create or merge .mcp.json with MCP server configuration
+#
+# A pre-existing .mcp.json (e.g. user-maintained `github` / `figma` entries)
+# must NOT be overwritten or skipped — we merge core entries in while
+# preserving everything the user added. See issue #315.
 # ---------------------------------------------------------------------------
 $mcpJsonPath = Join-Path $ProjectDir ".mcp.json"
-if (Test-Path $mcpJsonPath) {
-    Write-DotbotWarning ".mcp.json already exists -- skipping"
+
+# Playwright MCP output goes to .bot/.control/ (gitignored) — uses a relative
+# path so .mcp.json doesn't contain absolute user paths that trip the privacy scan
+$pwOutputDir = ".bot/.control/playwright-output"
+
+# On Windows, npx must be invoked via 'cmd /c' for stdio MCP servers
+if ($IsWindows) {
+    $npxCommand = "cmd"
+    $npxContext7Args = @("/c", "npx", "-y", "@upstash/context7-mcp@latest")
+    $npxPlaywrightArgs = @("/c", "npx", "-y", "@playwright/mcp@latest", "--output-dir", $pwOutputDir)
 } else {
-    Write-Status "Creating .mcp.json (dotbot + Context7 + Playwright)"
+    $npxCommand = "npx"
+    $npxContext7Args = @("-y", "@upstash/context7-mcp@latest")
+    $npxPlaywrightArgs = @("-y", "@playwright/mcp@latest", "--output-dir", $pwOutputDir)
+}
 
-    # Playwright MCP output goes to .bot/.control/ (gitignored) — uses a relative
-    # path so .mcp.json doesn't contain absolute user paths that trip the privacy scan
-    $pwOutputDir = ".bot/.control/playwright-output"
+# Core entries owned by the framework. Order matters — written in this order
+# in the resulting file, with any user-added entries appended after.
+$coreServers = [ordered]@{
+    dotbot = [ordered]@{
+        type    = "stdio"
+        command = "pwsh"
+        args    = @("-NoProfile", "-ExecutionPolicy", "Bypass", "-File", ".bot/core/mcp/dotbot-mcp.ps1")
+        env     = @{}
+    }
+    context7 = [ordered]@{
+        type    = "stdio"
+        command = $npxCommand
+        args    = $npxContext7Args
+        env     = @{}
+    }
+    playwright = [ordered]@{
+        type    = "stdio"
+        command = $npxCommand
+        args    = $npxPlaywrightArgs
+        env     = @{}
+    }
+}
 
-    # On Windows, npx must be invoked via 'cmd /c' for stdio MCP servers
-    if ($IsWindows) {
-        $npxCommand = "cmd"
-        $npxContext7Args = @("/c", "npx", "-y", "@upstash/context7-mcp@latest")
-        $npxPlaywrightArgs = @("/c", "npx", "-y", "@playwright/mcp@latest", "--output-dir", $pwOutputDir)
-    } else {
-        $npxCommand = "npx"
-        $npxContext7Args = @("-y", "@upstash/context7-mcp@latest")
-        $npxPlaywrightArgs = @("-y", "@playwright/mcp@latest", "--output-dir", $pwOutputDir)
+if (Test-Path $mcpJsonPath) {
+    Write-Status "Merging .mcp.json (preserving user entries)"
+    try {
+        $existing = Get-Content $mcpJsonPath -Raw | ConvertFrom-Json -ErrorAction Stop
+    } catch {
+        throw ".mcp.json exists but is not valid JSON: $($_.Exception.Message). Fix or remove the file and re-run dotbot init."
     }
 
-    $mcpConfig = @{
-        mcpServers = [ordered]@{
-            dotbot = [ordered]@{
-                type    = "stdio"
-                command = "pwsh"
-                args    = @("-NoProfile", "-ExecutionPolicy", "Bypass", "-File", ".bot/core/mcp/dotbot-mcp.ps1")
-                env     = @{}
-            }
-            context7 = [ordered]@{
-                type    = "stdio"
-                command = $npxCommand
-                args    = $npxContext7Args
-                env     = @{}
-            }
-            playwright = [ordered]@{
-                type    = "stdio"
-                command = $npxCommand
-                args    = $npxPlaywrightArgs
-                env     = @{}
+    $mergedServers = [ordered]@{}
+    $existingHasMcpServers = $existing -and $existing.PSObject.Properties['mcpServers'] -and $existing.mcpServers
+
+    # Core first, in canonical order
+    foreach ($coreName in $coreServers.Keys) {
+        $existingEntry = $null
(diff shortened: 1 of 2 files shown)


scripts/init-project.ps1
```diff
@@ -876,52 +876,103 @@ if (Test-Path $initScript) {
...
-                args    = $npxContext7Args
-                env     = @{}
-            }
-            playwright = [ordered]@{
-                type    = "stdio"
-                command = $npxCommand
-                args    = $npxPlaywrightArgs
-                env     = @{}
+    $mergedServers = [ordered]@{}
+    $existingHasMcpServers = $existing -and $existing.PSObject.Properties['mcpServers'] -and $existing.mcpServers
```
andresharpe: **[major] (80)** [PWSH-DIFF-MCPSERVERS-WRONG-TYPE]

Wrong-type `mcpServers` silently corrupts merged file. The guard `$existingHasMcpServers = $existing -and $existing.PSObject.Properties['mcpServers'] -and $existing.mcpServers` accepts an array, string, or scalar `mcpServers` as truthy. The downstream foreach over `$existing.mcpServers.PSObject.Properties` on a non-object yields the .NET intrinsic properties (Length, Rank, SyncRoot, IsReadOnly, IsFixedSize, IsSynchronized, Count, LongLength), which then get assigned into `$mergedServers` and serialized back to disk.

A user with `{"mcpServers": ["a","b"]}` (legal JSON, wrong shape) silently has init produce a merged file containing `"Length": 2`, `"Rank": 1`, `"SyncRoot": ["a","b"]`, etc. alongside the canonical core entries. Init reports success; the file is silently corrupted. Reproduced independently in PowerShell 7.

Fix: Verify object shape on the guard and throw a friendly error if the type is wrong, mirroring the pattern used for invalid JSON.

```suggestion
    $hasMcpServersProp = $existing -and $existing.PSObject.Properties['mcpServers']
    if ($hasMcpServersProp -and -not ($existing.mcpServers -is [System.Management.Automation.PSCustomObject])) {
        throw ".mcp.json has 'mcpServers' but it is not an object: got $($existing.mcpServers.GetType().Name). Fix or remove the file and re-run dotbot init."
    }
    $existingHasMcpServers = $hasMcpServersProp -and $existing.mcpServers
```

tests/Test-Structure.ps1
```diff
@@ -379,6 +379,101 @@ if (-not $dotbotInstalled) {
...
+        $invalidOutput = & pwsh -NoProfile -ExecutionPolicy Bypass -File (Join-Path $dotbotDir "scripts\init-project.ps1") 2>&1 | Out-String
+        $invalidExit = $LASTEXITCODE
+        Pop-Location
+
+        Assert-True -Name "merge: invalid JSON fails loudly (non-zero exit AND specific error)" `
+            -Condition (($invalidExit -ne 0) -and ($invalidOutput -match '(?i)(not valid json|invalid json|convertfrom-json|unexpected character)')) `
+            -Message "Expected init to fail with non-zero exit and invalid-JSON error for .mcp.json, got exit=$invalidExit, output: $invalidOutput"
+    } finally {
+        Remove-TestProject -Path $invalidProject
+    }
```
andresharpe, line 476: **[praise] (90)** [PWSH-PRAISE]

Well-shaped invalid-JSON regression test. Case 4 combines `($invalidExit -ne 0) -and ($invalidOutput -match '(?i)(not valid json|invalid json|convertfrom-json|unexpected character)')` -- non-zero exit AND a regex on the actual error text. Exactly the pattern that catches "init silently swallows the failure" regressions. The PR's response to Copilot's prior tightening request lands cleanly.

andresharpe (changes requested):
## Code review summary

_pwsh-code-review v0.19.0_

0 blocker, 1 major, 0 minor, 0 nit, 0 question

Verdict: **fix majors first**

---

andresharpe:
On the two open Copilot threads about `-Force` semantics for core entries:

The original behaviour (preserve core entries unless `-Force`) is what reintroduces the exact downstream symptom this PR set out to fix — `dotbot init` reporting success while `.bot/go.ps1` still fails with "Dotbot MCP server not registered" — for any project that ran init before a framework path move (e.g. #345 moving the dotbot MCP server from `.bot/systems/` to `.bot/core/`). The user has to know to run `dotbot init -Force` to recover, which they generally don't.

I went with the framework-ownership model: core entries (dotbot, context7, playwright) are always set to the canonical value; user-added entries with non-core names are preserved verbatim. This matches the rest of dotbot's posture toward `.bot/` content and the new test case 7 locks the self-heal in.

If a user genuinely wants a customised MCP server, the right path is to add it under a non-core name rather than tampering with a framework-owned entry. Happy to revisit if usage data shows people relying on the old preserve-by-default behaviour.

andresharpe (approved):
Final /pwsh-review pass: address one [major] mcpServers wrong-type guard, plus three further majors surfaced by the multi-agent review (null root crash, non-object root corruption, stale-core preservation reintroducing the original bug). Added 3 regression tests (empty file, non-object root, stale-path self-heal) plus a case-insensitive name match for core servers. CI green across all platforms; all threads resolved. See PR comment for the design rationale on always-refreshing core entries.

PR #387, opened 2026-04-29
fix(runtime): split intentional skips from framework errors (#318)

### LINKED ISSUE #318: task_mark_skipped is overloaded for "not applicable" cases, causing retry loops, worktree collisions, and permanent dependent-task deadlocks
### What happened

`task_mark_skipped` is the only terminal-state tool an agent has for saying "this task should not be executed" — for instance, a pre-flight check that determines the task is not applicable (e.g. "the issue is missing the `needs-design` label, so skip this phase"). Agents reach for it in these situations. But the tool and the surrounding state machine treat skipping as a framework-level error outcome, not an intentional control-flow decision, and the effects compound:

1. **Schema/semantic overload.** `workflows/default/systems/mcp/tools/task-mark-skipped/metadata.yaml:12-14` restricts `skip_reason` to `non-recoverable | max-retries`. The runtime enforces the same values at `task-mark-skipped/script.ps1:14-17`. These are error terminals, but the tool is exposed to agents and there is no sibling tool for "intentional not-applicable". Agents therefore pick whichever reason looks least wrong and call this tool.

2. **Completion check does not treat `skipped/` as terminal.** `workflows/default/systems/runtime/modules/test-task-completion.ps1:33` only checks `Test-TaskDone`. When the task JSON lands in `skipped/`, `Test-TaskCompletion` returns `completed=false`, and the task-runner at `workflows/default/systems/runtime/modules/ProcessTypes/Invoke-WorkflowProcess.ps1:1048` logs `"Completion check failed ... not found in in-progress/ or done/ (unexpected state)"` and enters the retry branch.

3. **Retry causes a worktree collision.** `task_mark_skipped` does not clean up the per-task worktree — it only moves the task JSON and writes `skip_history`. On retry, `New-TaskWorktree` calls `git worktree add -b <branch> <path> <base>` at `WorktreeManager.psm1:505`, which fails because the path and branch are already registered. The retry path throws. After `max_retries_per_task` attempts, the loop finally calls `task_mark_skipped` itself with `skip_reason=max-retries` — so the task ends up in `skipped/` for the opposite of the original reason.

4. **No `failed` state; `skipped` carries both meanings.** `workflows/default/systems/mcp/modules/TaskStore.psm1:16-17` lists the valid statuses as `todo, analysing, needs-input, analysed, in-progress, done, split, skipped, cancelled`. There is no `failed`. Intentional skips and terminal errors both land in `skipped/`, which makes downstream behavior ambiguous.

5. **Dependent tasks deadlock.** `workflows/default/systems/mcp/modules/TaskIndexCache.psm1:706-723` (`Test-AllDependenciesMet`) only satisfies dependencies from the `Done` set. A task with a `skipped` prerequisite is permanently ineligible — it stays in `analysed/` forever while `task-get-next` reports `"N analysed task(s) blocked by unmet dependencies"`.

The result is that a single agent-initiated "not applicable" skip cascades into: retry loop → worktree collision → max-retries skip → blocked dependents → workflow stall. None of the surface symptoms point back to the root cause, which is that `skipped` is being asked to mean three different things at once.

### What you expected

The state machine should distinguish between three distinct terminal outcomes, and each MCP tool/schema should correspond to exactly one. Minimum set of fixes:

1. **Treat `skipped/` and `cancelled/` as terminal in the completion check.** `Test-TaskCompletion` should return `completed=true` (with a distinguishing `reason`/`method`) when the task JSON is found in any terminal directory. The retry branch in `Invoke-WorkflowProcess.ps1` should not fire when a task is already in a terminal state.

2. **Introduce a `failed` state (or equivalent separation).** Framework-level unrecoverable errors and max-retries outcomes should move the task to `failed/`, not `skipped/`. `TaskStore.psm1` should add `failed` to the valid-statuses list and `task_mark_skipped` should stop accepting `non-recoverable` / `max-retries` — those belong on a separate tool (`task_mark_failed`) or on a typed argument that makes the distinction explicit.

3. **Constrain `task_mark_skipped` to intentional skips only.** Its `skip_reason` enum should change to describe intentional-skip reasons (e.g. `not-applicable`, `precondition-unmet`, user-supplied free text), not error terminals. This is the tool agents should keep reaching for, with semantics that match how they reach for it.

4. **Ensure the terminating tools clean up their worktree.** Whichever tool(s) move a task into a terminal state should either remove the per-task worktree or leave it in a state that `New-TaskWorktree` can safely re-enter on the next iteration. No path where a terminal state and a live worktree coexist.

5. **Decide how dependents handle non-done terminals.** At least two options:
   - Treat `skipped` (intentional) as a dependency-satisfier — the dependent task runs. `failed` still blocks.
   - Keep both as blockers but surface the block with a clear status (`blocked-by-skipped`, `blocked-by-failed`) rather than leaving the task silently in `analysed/`.

   Either is defensible; the current silent deadlock is not.

After the fix, an agent calling `task_mark_skipped` for a "not applicable" pre-flight result should produce: task in a terminal state, no retry, no worktree collision, and a deterministic outcome for downstream tasks (either eligible or clearly blocked with a reason).

### Steps to reproduce

1. In a clean test repo, run `dotbot init`.
2. Create two tasks with a dependency between them. `TaskB` depends on `TaskA`:
   ```
   dotbot task create --name "TaskA" --description "Pre-flight check: call task_mark_skipped with skip_reason='non-recoverable' to simulate a 'not applicable' outcome, then stop."
   dotbot task create --name "TaskB" --description "Do something trivial" --depends-on TaskA
   ```
   (The description for `TaskA` is what makes the agent pick up `task_mark_skipped` — the default prompts do not instruct this, so the test prompt has to. Alternatively, call `task_mark_skipped` directly via the MCP tool against a claimed task JSON.)
3. Start autonomous execution.
4. Observe the following in the task-runner log:
   - `TaskA` JSON moves to `.bot/workspace/tasks/skipped/`.
   - The task-runner logs `"Completion check failed ... not found in in-progress/ or done/ (unexpected state)"` for `TaskA`.
   - The task-runner retries. `New-TaskWorktree` fails with a git error along the lines of `fatal: '<path>' is already used by worktree at <path>` (or an equivalent "branch already checked out" message).
   - After `max_retries_per_task` iterations, the runtime itself calls `task_mark_skipped` with `skip_reason=max-retries` — so `TaskA` now has two entries in `skip_history`, the second one labeled as a framework error.
5. Inspect `TaskB`:
   ```
   dotbot task get-next
   ```
   Observed: `TaskB` reports as blocked (`"N analysed task(s) blocked by unmet dependencies"`) and stays in `analysed/` indefinitely, because `Test-AllDependenciesMet` only considers `Done` to satisfy a dependency. The workflow stalls.

Expected after the fix: in step 4, `TaskA` enters a terminal state on the first call, no retry, no worktree collision, and no framework-initiated `max-retries` skip. In step 5, `TaskB` either runs (if intentional skips are treated as satisfying dependencies) or is surfaced as explicitly `blocked-by-skipped` with a clear reason, not silently stuck.

### Environment

All environments

### Severity

medium

### Logs / screenshots

_No response_

kabaogluemre:
## Summary

Closes #318 — `task_mark_skipped` was overloaded for both intentional "not applicable" decisions and framework errors. The result: completion check missed `skipped/` as terminal → retry loop → worktree collision → max-retries skip → silently deadlocked dependents. None of the symptoms pointed back to the root cause.

This PR splits the two meanings cleanly via `skip_history[].reason` discrimination (no new physical state directory — kept the storage simple per the issue's "or equivalent separation" allowance).

### State machine
- **`Test-TaskCompletion`** now returns `completed=true` with `method=TerminalState` for any task that lands in `skipped/`, `cancelled/`, or `split/`. The runner breaks out of the retry loop and cleans up the worktree **without squash-merging** — so an agent calling `task_mark_skipped` mid-execution no longer silently merges its abandoned work to `main`.
- **`Get-TaskTerminalState`** added to `TaskIndexCache.psm1` as the single check for terminal residency.

### Skip-reason classification
Two disjoint sets, both validated by the agent-facing tool:

| Class | Reasons | Effect on dependents | Auto-retried by `Reset-SkippedTasks`? |
|---|---|---|---|
| **Intentional** | `not-applicable`, `precondition-unmet`, `superseded`, `user-requested`, `condition-not-met` | Satisfy (added to `DoneIds`) | No — deliberate decision, leave it alone |
| **Framework error** | `non-recoverable`, `max-retries` | Block (surfaced via `Get-DeadlockedTasks` with explicit blocker name) | Yes, until `skip_history.Count >= 3` |

`task_mark_skipped` accepts both, but the metadata description steers agents to intentional reasons; the runtime emits framework reasons itself.

### Single source of truth
The reason lists used to drift across three files. Now consolidated in `core/mcp/modules/TaskIndexCache.psm1`:

- `Get-IntentionalSkipReasons`, `Get-FrameworkSkipReasons` — exported accessors
- `Test-IsFrameworkErrorSkip -TaskContent <obj>` — canonical classifier (reads `skip_history[-1].reason`, falls back to top-level `skip_reason` for `task-get-next` callers)

`task-mark-skipped/script.ps1` and `core/runtime/modules/task-reset.ps1` import from there instead of inlining.

### Worktree cleanup
`Invoke-WorkflowProcess.ps1` now has a dedicated terminal-state branch: when a task settles into `skipped`/`cancelled`/`split`, the worktree is removed and the worktree map updated, but `consecutive_failures` is **not** bumped (these aren't failures) and no merge is attempted.

## Files touched

| File | Why |
|---|---|
| `core/mcp/modules/TaskIndexCache.psm1` | Canonical reason lists, `Test-IsFrameworkErrorSkip`, `Get-TaskTerminalState`; intentional skips added to `DoneIds`; `Get-DeadlockedTasks` filters to framework-only blockers |
| `core/mcp/tools/task-mark-skipped/{metadata.yaml,script.ps1}` | Enum expanded with intentional reasons, optional `skip_detail` field, `intentional` flag in response |
| `core/runtime/modules/test-task-completion.ps1` | Terminal-state branch using `Get-TaskTerminalState` |
| `core/runtime/modules/task-reset.ps1` | `Reset-SkippedTasks` only auto-retries framework-error skips |
| `core/runtime/modules/ProcessTypes/Invoke-WorkflowProcess.ps1` | New terminal-state cleanup branch; runtime calls to `Invoke-TaskMarkSkipped` use proper enum values + `skip_detail` |
| `tests/Test-TaskActions.ps1` | New `Test-TaskCompletion` terminal-state suite (6 scenarios), new `Reset-SkippedTasks` polarity guard suite (4 scenarios), `Get-DeadlockedTasks` updated for intentional-vs-framework |
| `core/mcp/tools/task-mark-skipped/test.ps1` | Asserts `intentional` flag, accepts framework reasons, rejects unknown reasons |

## Reproduction (from issue #318) — before/after

| Step | Before | After |
|---|---|---|
| Agent calls `task_mark_skipped` mid-execution | `Test-TaskCompletion` returns `completed=false` → retry → `New-TaskWorktree` fails (path/branch already registered) → max-retries skip appended | `Test-TaskCompletion` returns `completed=true method=TerminalState` → runner breaks → worktree removed cleanly |
| Dependent of an intentionally skipped task | Stuck in `analysed/` indefinitely; `task-get-next` reports "blocked by unmet dependencies" with no detail | Runs (intentional skip is in `DoneIds`) |
| Dependent of a framework-error skipped task | Same silent deadlock | Surfaced by `Get-DeadlockedTasks` with explicit blocker name |

## Test plan

- [x] `pwsh tests/Run-Tests.ps1` — Layer 2 + Layer 3 green; Test-TaskActions covers all new branches
- [x] `Test-TaskCompletion` returns correct `method`/`terminal_state` for done / intentional-skip / framework-skip / cancelled / split / in-progress
- [x] `Get-DeadlockedTasks` reports zero blockers when prereq is intentionally skipped, reports framework-error blocker by name otherwise
- [x] `Reset-SkippedTasks` leaves intentional skips, retries framework skips, leaves `skip_count >= 3` for manual review, handles top-level `skip_reason` fallback (task-get-next path)
- [x] `task-mark-skipped` accepts intentional + framework reasons, rejects unknown, returns `intentional` flag

## Notes

- No new physical state directory; the issue's "or equivalent separation" allowance covered this. If a `failed/` directory is preferred later, the discriminator function is the only place to swap.
- Layer 1 `Test-Structure.ps1` failure observed during local run was an environmental Windows file-lock on the user's `~/dotbot` install dir, unrelated to this change.

### DIFF
--- a/core/runtime/modules/ProcessTypes/Invoke-WorkflowProcess.ps1
+++ b/core/runtime/modules/ProcessTypes/Invoke-WorkflowProcess.ps1
@@ -1629,7 +1629,7 @@ Work on this task autonomously. When complete, ensure you call task_mark_done vi
             if (-not $failureReason.recoverable) {
                 Write-Status "Non-recoverable failure - skipping" -Type Error
                 try {
-                    $detail = $failureReason.reason ?? $failureReason.message ?? 'non-recoverable failure'
+                    $detail = $failureReason.description ?? $failureReason.type ?? 'non-recoverable failure'
                     Invoke-TaskMarkSkipped -Arguments @{ task_id = $task.id; skip_reason = 'non-recoverable'; skip_detail = $detail } | Out-Null
                 } catch { Write-BotLog -Level Warn -Message "Task operation failed" -Exception $_ }
                 break
--- a/core/runtime/modules/ProcessTypes/Invoke-WorkflowProcess.ps1
+++ b/core/runtime/modules/ProcessTypes/Invoke-WorkflowProcess.ps1
@@ -835,7 +835,7 @@ try {
                     Write-Status $typeError -Type Error
                     Write-ProcessActivity -Id $procId -ActivityType "error" -Message "$($task.name): $typeError"
                     try {
-                        Invoke-TaskMarkSkipped -Arguments @{ task_id = $task.id; skip_reason = $typeError } | Out-Null
+                        Invoke-TaskMarkSkipped -Arguments @{ task_id = $task.id; skip_reason = 'non-recoverable'; skip_detail = $typeError } | Out-Null
                     } catch { Write-BotLog -Level Debug -Message "Logging operation failed" -Exception $_ }
                     if (Test-TaskIsMandatory $task) {
                         Write-Status "Mandatory task failed: $($task.name) - stopping workflow" -Type Error
@@ -876,7 +876,7 @@ try {
                     Write-Status $typeError -Type Error
                     Write-ProcessActivity -Id $procId -ActivityType "error" -Message "$($task.name): $typeError"
                     try {
-                        Invoke-TaskMarkSkipped -Arguments @{ task_id = $task.id; skip_reason = $typeError } | Out-Null
+                        Invoke-TaskMarkSkipped -Arguments @{ task_id = $task.id; skip_reason = 'non-recoverable'; skip_detail = $typeError } | Out-Null
                     } catch { Write-BotLog -Level Debug -Message "Logging operation failed" -Exception $_ }
                     if (Test-TaskIsMandatory $task) {
                         Write-Status "Mandatory task failed: $($task.name) - stopping workflow" -Type Error
@@ -1100,7 +1100,7 @@ try {
             } else {
                 Write-Status "Task failed: $($task.name)" -Type Error
                 try {
-                    Invoke-TaskMarkSkipped -Arguments @{ task_id = $task.id; skip_reason = "$taskTypeVal execution failed: $typeError" } | Out-Null
+                    Invoke-TaskMarkSkipped -Arguments @{ task_id = $task.id; skip_reason = 'non-recoverable'; skip_detail = "$taskTypeVal execution failed: $typeError" } | Out-Null
                 } catch { Write-BotLog -Level Debug -Message "Session operation failed" -Exception $_ }
 
                 # Mandatory-task halt (#213): script/mcp/task_gen failure
@@ -1482,6 +1482,12 @@ Work on this task autonomously. When complete, ensure you call task_mark_done vi
         # — its worktree must be retained so the executor can resume after
         # task_answer_question moves the task back to analysing/.
         $taskParked = $false
+        # Set when the task ended in a terminal state other than done
+        # (skipped/cancelled/split). Distinct from taskSuccess because we must
+        # NOT squash-merge the worktree, NOT count the task as completed, and
+        # NOT log "task -> done". The worktree still has to be cleaned up.
+        $taskTerminal = $false
+        $taskTerminalState = $null
         $postScriptFailed = $false
         $postScriptError = $null
         # Distinguishes which post-task hook actually flipped postScriptFailed
@@ -1557,8 +1563,21 @@ Work on this task autonomously. When complete, ensure you call task_mark_done vi
 
             # Check completion
             $completionCheck = Test-TaskCompletion -TaskId $task.id
-            Write-Diag "Completion check: completed=$($completionCheck.completed)"
+            Write-Diag "Completion check: completed=$($completionCheck.completed) method=$($completionCheck.method) terminal_state=$($completionCheck.terminal_state)"
             if ($completionCheck.completed) {
+                # Issue #318: distinguish done from other terminal states
+                # (skipped/cancelled/split). Only done squash-merges to main and
+                # counts as a completed task. Other terminals must clean up the
+                # worktree without merging — otherwise an agent calling
+                # task_mark_skipped silently merges its abandoned work.
+                if ($completionCheck.method -eq 'TerminalState' -and $completionCheck.terminal_state -ne 'done') {
+                    $taskTerminalState = $completionCheck.terminal_state
+                    Write-Status "Task ended in terminal state: $taskTerminalState" -Type Info
+                    Write-Information "task_state_change: $($task.id) -> $taskTerminalState [execution]" -Tags @('dotbot', 'task', 'state')
+                    Write-ProcessActivity -Id $procId -ActivityType "text" -Message "Task ended in terminal state '$taskTerminalState': $($task.name)"
+                    $taskTerminal = $true
+                    break
+                }
                 Write-Status "Task completed!" -Type Complete
                 Write-Information "task_state_change: $($task.id) -> done [execution]" -Tags @('dotbot', 'task', 'state')
                 Invoke-SessionIncrementCompleted -Arguments @{} | Out-Null
@@ -1610,7 +1629,8 @@ Work on this task autonomously. When complete, ensure you call task_mark_done vi
             if (-not $failureReason.recoverable) {
                 Write-Status "Non-recoverable failure - skipping" -Type Error
                 try {
-                    Invoke-TaskMarkSkipped -Arguments @{ task_id = $task.id; skip_reason = "non-recoverable" } | Out-Null
+                    $detail = $failureReason.reason ?? $failureReason.message ?? 'non-recoverable failure'
+                    Invoke-TaskMarkSkipped -Arguments @{ task_id = $task.id; skip_reason = 'non-recoverable'; skip_detail = $detail } | Out-Null
                 } catch { Write-BotLog -Level Warn -Message "Task operation failed" -Exception $_ }
                 break
             }
@@ -1618,7 +1638,7 @@ Work on this task autonomously. When complete, ensure you call task_mark_done vi
             if ($attemptNumber -ge $maxRetriesPerTask) {
                 Write-Status "Max retries exhausted" -Type Error
                 try {
-                    Invoke-TaskMarkSkipped -Arguments @{ task_id = $task.id; skip_reason = "max-retries" } | Out-Null
+                    Invoke-TaskMarkSkipped -Arguments @{ task_id = $task.id; skip_reason = 'max-retries'; skip_detail = "Retry budget exhausted after $attemptNumber attempt(s)" } | Out-Null
                 } catch { Write-BotLog -Level Warn -Message "Task operation failed" -Exception $_ }
                 break
             }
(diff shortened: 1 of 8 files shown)


core/runtime/modules/ProcessTypes/Invoke-WorkflowProcess.ps1
```diff
@@ -1610,15 +1629,16 @@ Work on this task autonomously. When complete, ensure you call task_mark_done vi
             if (-not $failureReason.recoverable) {
                 Write-Status "Non-recoverable failure - skipping" -Type Error
                 try {
-                    Invoke-TaskMarkSkipped -Arguments @{ task_id = $task.id; skip_reason = "non-recoverable" } | Out-Null
+                    $detail = $failureReason.reason ?? $failureReason.message ?? 'non-recoverable failure'
```
andresharpe: **[major] (85)** [PWSH-LANG-NULL-COALESCE-MISSING-KEY]

The new `??` chain `$failureReason.reason ?? $failureReason.message ?? 'non-recoverable failure'` reads two keys (`reason`, `message`) that the source hashtable does not expose. `Get-FailureReason` (core/runtime/modules/get-failure-reason.ps1:34-118) returns hashtables whose keys are `type`, `description`, `recoverable`, `suggested_action`. Under the project's `Set-StrictMode -Version 1.0`, the missing-key access silently returns $null, so both `??` operands evaluate null and `$detail` always collapses to the literal string `'non-recoverable failure'`.

The new `skip_detail` field never carries the actual failure description that the diff intends to surface. Operators inspecting `skipped/` to triage non-recoverable failures see only the boilerplate `'non-recoverable failure'`, defeating the new audit-trail discriminator. The other two new `Invoke-TaskMarkSkipped` callers (lines 838, 879, type-error paths) and the max-retries call at 1641 all pass meaningful `skip_detail`; this branch alone regresses.

Fix: Use the keys that `Get-FailureReason` actually returns. `description` carries the human-readable explanation; `type` carries the classification token.

```suggestion
                    $detail = $failureReason.description ?? $failureReason.type ?? 'non-recoverable failure'
                    Invoke-TaskMarkSkipped -Arguments @{ task_id = $task.id; skip_reason = 'non-recoverable'; skip_detail = $detail } | Out-Null
```

tests/Test-TaskActions.ps1
```diff
@@ -757,6 +787,274 @@ finally {
...
+
+    Initialize-TaskIndex -TasksBaseDir $tasksBaseDir
+    $resultFramework = Test-TaskCompletion -TaskId "tc-framework"
+    Assert-True -Name "Framework-error skip is reported completed=true (issue #318)" `
+        -Condition ($resultFramework.completed -eq $true) `
+        -Message "Expected completed=true, got $($resultFramework.completed)"
+    Assert-Equal -Name "Framework-error skip reports method=TerminalState" `
+        -Expected "TerminalState" -Actual $resultFramework.method
+    Assert-Equal -Name "Framework-error skip reports terminal_state=skipped" `
+        -Expected "skipped" -Actual $resultFramework.terminal_state
```
andresharpe, line 880: **[praise] (90)** [PWSH-PRAISE]

Test coverage is the strongest part of this PR. +322 lines covering six terminal-state scenarios in `Test-TaskCompletion` (done / intentional-skip / framework-skip / cancelled / split / in-progress) plus four polarity guards on `Reset-SkippedTasks` (intentional left alone, framework retried, `>= 3` left for manual review, top-level `skip_reason` fallback). The polarity assertions in particular are exactly the regression guards that would have caught the original #318 bug if they had existed earlier. Worth modeling for future state-machine refactors.

andresharpe (changes requested):
## Code review summary

_pwsh-code-review v0.19.0_

0 blocker, 1 major, 0 minor, 0 nit, 0 question

Verdict: **fix majors first**

---

andresharpe (approved):
Final /pwsh-review pass: 0 in-scope findings from all 5 agents. Major key-mismatch fix verified against Get-FailureReason's actual return shape (description, type). Two follow-up Copilot nits addressed: cache is_framework_error_skip on index entries (no second filesystem pass in Get-DeadlockedTasks) and surface a clear error if Test-IsFrameworkErrorSkip cannot be loaded in Reset-SkippedTasks. CI green across all platforms. All threads resolved.

PR #409, opened 2026-05-09
fix(workflow-manifest): loud schema errors for malformed env_vars (#319)

### LINKED ISSUE #319: env_vars schema: null crash on one path, silent skip on the other — every new workflow author loses time here
### What happened

The `requires.env_vars` schema in workflow manifests expects entries shaped as `{ var, name, message, hint }`, where:

- `var` is the environment variable identifier (e.g. `GITHUB_TOKEN`)
- `name` is the human-readable display name (e.g. `"GitHub Personal Access Token"`)

A workflow author writing their first manifest naturally types `name: GITHUB_TOKEN` — it is the obvious choice without prior exposure to this schema, and there is no authoritative reference to correct the assumption upfront. There is no validator that flags the mistake, no schema error, no hint in the docs path most authors encounter first.

Two code paths then diverge into two different bad failure modes, both traceable to the same root cause (`var` read without a fallback and without a null guard):

**1. Crash path — `New-EnvLocalScaffold`** (`workflows/default/systems/runtime/modules/workflow-manifest.ps1:509-521`):

```powershell
foreach ($ev in $EnvVars) {
    $varName = $ev.var
    if (-not $varName) { $varName = $ev['var'] }
    ...
    if ($existing.ContainsKey($varName)) { ... }
}
```

When `var` is missing, `$varName` is `$null`, and line 515 calls `ContainsKey($null)`, which throws:

```
Exception calling "ContainsKey" with "1" argument(s): "Value cannot be null. (Parameter 'key')"
```

The error names no entry, no field, no workflow file. An author with a 10-line workflow can debug it in a minute; an author with a richer manifest has to diff against an example to find the typo.

**2. Silent-skip path — requires-checks builder** (`workflows/default/systems/runtime/modules/workflow-manifest.ps1:131-140`):

```powershell
$varName = if ($ev -is [System.Collections.IDictionary]) { $ev['var'] } else { $ev.var }
...
if ($varName) {
    $checks += @{ type = 'env_var'; var = $varName; name = ...; message = ...; hint = ... }
}
```

Here, if `var` is missing, the entry is silently dropped from the checks list. The author who uses `name: GITHUB_TOKEN` sees no crash — but the env var is never validated. They discover the problem later, indirectly, when the agent fails auth or the MCP call 401s. This is worse than the crash: the crash at least points at a specific code path.

**DX cost:** this is a silent tax on every new workflow author. The first thing an author does is write `requires.env_vars`, and the first thing the framework does is either crash without pointing at the cause or validate nothing at all without saying so. The fix is trivial; the ongoing cost of leaving it is not.

### What you expected

Either path, hit with a malformed `env_vars` entry, should produce an actionable error that tells the author exactly what is wrong and exactly how to fix it. Concretely:

1. **Both paths** (`New-EnvLocalScaffold` at line 510 and the requires-checks builder at line 135) should replace silent nulls / silent skips with a loud error like:

   ```
   env_vars entry in workflow '<name>' is missing the required 'var' field.
   Entry: { name: "GITHUB_TOKEN", message: "required" }
   Expected schema: { var: <IDENTIFIER>, name: <DISPLAY NAME>, message: <TEXT>, hint: <TEXT> }
   Note: 'var' is the env var identifier (e.g. GITHUB_TOKEN). 'name' is the human-readable label (e.g. "GitHub Personal Access Token").
   ```

2. **Validate at install time, not at runtime.** `dotbot init --workflow <name>` and `dotbot workflow add <name>` should run a schema check over the parsed manifest before copying files, so authors get the error at the point they can act on it, not deep into an autonomous run. A `Test-WorkflowManifestSchema` helper in `workflow-manifest.ps1` that is called from both entry points covers it.

3. **Do not "fix" the crash by falling back to `ev.name`.** `var` and `name` are deliberately distinct fields with different meanings. Silently treating one as the other would hide schema errors and later produce `.env.local` files with odd display-label identifiers. The right fix is the clear error, not the implicit coercion.

### Steps to reproduce

1. In a clean test repo, create a minimal workflow at `~/dotbot/workflows/my-test-workflow/workflow.yaml`:

   ```yaml
   name: my-test-workflow
   version: "1.0"
   description: "Repro for env_vars schema error"
   min_dotbot_version: "4.0"

   requires:
     env_vars:
       - name: GITHUB_TOKEN
         message: "required"
   ```

   Note the entry uses `name:` instead of `var:` — the natural choice for a first-time author.

2. Run `dotbot init --workflow my-test-workflow -Force` in the test repo.

3. Observed, via the crash path (`New-EnvLocalScaffold`):

   ```
   Exception calling "ContainsKey" with "1" argument(s): "Value cannot be null. (Parameter 'key')"
   ```

   No indication of which workflow, which entry, or which field is wrong. `.env.local` is not created correctly.

4. Alternatively, to hit the silent-skip path, run the requires-checks logic that consumes the same manifest (e.g. via the pre-flight `requires` check flow). Observed: no error, no warning — the `GITHUB_TOKEN` check is quietly absent from the resulting checks list. At runtime the agent later fails with a 401/auth error from whatever tool expected the variable.

Expected after the fix: step 3 produces an error that names the workflow, quotes the offending entry, states that `var` is required, and shows the correct schema. Step 4 emits the same error rather than silently dropping the entry.

### Environment

All environments

### Severity

low

### Logs / screenshots

_No response_

kabaogluemre:
Closes #319.

`requires.env_vars` entries missing a `var` field hit two different bad failure modes — both fixed:

1. **Crash path** — `New-EnvLocalScaffold` called `Hashtable.ContainsKey($null)` and threw `Value cannot be null` with no workflow, entry, or field info.
2. **Silent-skip path** — `Convert-ManifestRequiresToPreflightChecks` quietly dropped the entry, so the env var was never validated and the agent later 401'd at runtime.

## Changes

- **New `Test-WorkflowManifestSchema` helper** (`core/runtime/modules/workflow-manifest.ps1`) — validates `requires.env_vars` (must have `var`), `requires.mcp_servers` (must have `name`), `requires.cli_tools` (must have `name`). Returns one human-readable error per malformed entry, naming the workflow, the offending entry contents, and the expected schema. Includes the note that `var` is the identifier vs. `name` is the display label, since that's the most common author confusion.
- **Install-time gate** — `dotbot init` (`scripts/init-project.ps1`) and `dotbot workflow add` (`scripts/workflow-add.ps1`) run the validator after parsing the manifest and halt before any scaffolding, so authors see the error at the point they can fix it instead of deep into a run.
- **Defense-in-depth throws** — `New-EnvLocalScaffold` and `Convert-ManifestRequiresToPreflightChecks` now throw a clear schema error on malformed entries instead of crashing or silently dropping. `WorkflowName` flows through both for traceability.
- **Dashboard graceful handling** — `Get-PreflightResults` (`core/ui/modules/ProductAPI.psm1`) catches the throw and surfaces it as a single failed `manifest_schema` check rather than crashing the Product tab render.
- **No implicit fallback to `ev.name`** — per the issue, treating one field as the other would hide schema errors and produce odd identifiers in `.env.local`. The fix is the loud error, not coercion.

## Tests

`tests/Test-WorkflowManifest.ps1`:
- `Test-WorkflowManifestSchema`: valid manifest, no `requires`, missing `var`, multiple bad entries across all three sections, `WorkflowName` override
- `Convert-ManifestRequiresToPreflightChecks`: throws on missing `var` (env_vars), missing `name` (mcp_servers, cli_tools), with workflow and field named in the message
- `New-EnvLocalScaffold`: throws on missing `var` with the full schema-error format, and explicitly asserts the legacy `Value cannot be null` crash is no longer hit

### DIFF
--- a/core/runtime/modules/workflow-manifest.ps1
+++ b/core/runtime/modules/workflow-manifest.ps1
@@ -189,6 +189,130 @@ function Get-ActiveWorkflowManifest {
     return $null
 }
 
+function Get-ManifestEntryField {
+    param([object]$Entry, [string]$Field)
+    if ($null -eq $Entry) { return $null }
+    if ($Entry -is [System.Collections.IDictionary]) { return $Entry[$Field] }
+    return $Entry.$Field
+}
+
+function Format-ManifestEntryForError {
+    <#
+    .SYNOPSIS
+    Render a manifest entry as a compact "{ key: value, ... }" string for error messages.
+    #>
+    param([object]$Entry)
+    if ($null -eq $Entry) { return '<null>' }
+    if ($Entry -is [System.Collections.IDictionary]) {
+        $pairs = @()
+        foreach ($k in $Entry.Keys) {
+            $v = $Entry[$k]
+            $vRendered = if ($null -eq $v) { 'null' } elseif ($v -is [string]) { '"' + $v + '"' } else { [string]$v }
+            $pairs += "$k`: $vRendered"
+        }
+        return '{ ' + ($pairs -join ', ') + ' }'
+    }
+    $pairs = @()
+    foreach ($p in $Entry.PSObject.Properties) {
+        $vRendered = if ($null -eq $p.Value) { 'null' } elseif ($p.Value -is [string]) { '"' + $p.Value + '"' } else { [string]$p.Value }
+        $pairs += "$($p.Name): $vRendered"
+    }
+    return '{ ' + ($pairs -join ', ') + ' }'
+}
+
+function Test-WorkflowManifestSchema {
+    <#
+    .SYNOPSIS
+    Validate a parsed workflow manifest against the requires.* schema.
+
+    .DESCRIPTION
+    Returns an array of human-readable error strings — one per malformed entry.
+    Empty array means the manifest is valid for the requires.* sections.
+
+    Validates that every entry in:
+      - requires.env_vars     has a non-empty 'var' field
+      - requires.mcp_servers  has a non-empty 'name' field
+      - requires.cli_tools    has a non-empty 'name' field
+
+    Used at install time by `dotbot init` and `dotbot workflow add` to surface
+    schema mistakes before any scaffolding runs, so the author gets a clear
+    error at the point they can act on it instead of a null-key crash from
+    New-EnvLocalScaffold or a silently-dropped preflight check at runtime.
+    #>
+    param(
+        [Parameter(Mandatory)]
+        [object]$Manifest,
+
+        [string]$WorkflowName
+    )
+
+    $errors = @()
+    if (-not $WorkflowName) {
+        $WorkflowName = Get-ManifestEntryField -Entry $Manifest -Field 'name'
+        if (-not $WorkflowName) { $WorkflowName = '<unknown>' }
+    }
+
+    $requires = Get-ManifestEntryField -Entry $Manifest -Field 'requires'
+    if (-not $requires) { return @() }
+
+    # env_vars: each entry must have 'var'
+    $envVars = Get-ManifestEntryField -Entry $requires -Field 'env_vars'
+    if ($envVars) {
+        $i = 0
+        foreach ($ev in @($envVars)) {
+            $varName = Get-ManifestEntryField -Entry $ev -Field 'var'
+            if (-not $varName) {
+                $rendered = Format-ManifestEntryForError -Entry $ev
+                $errors += @"
+env_vars entry [$i] in workflow '$WorkflowName' is missing the required 'var' field.
+Entry: $rendered
+Expected schema: { var: <IDENTIFIER>, name: <DISPLAY NAME>, message: <TEXT>, hint: <TEXT> }
+Note: 'var' is the env var identifier (e.g. GITHUB_TOKEN). 'name' is the human-readable label (e.g. "GitHub Personal Access Token").
+"@
+            }
+            $i++
+        }
+    }
+
+    # mcp_servers: each entry must have 'name'
+    $mcpServers = Get-ManifestEntryField -Entry $requires -Field 'mcp_servers'
+    if ($mcpServers) {
+        $i = 0
+        foreach ($ms in @($mcpServers)) {
+            $srvName = Get-ManifestEntryField -Entry $ms -Field 'name'
+            if (-not $srvName) {
+                $rendered = Format-ManifestEntryForError -Entry $ms
+                $errors += @"
+mcp_servers entry [$i] in workflow '$WorkflowName' is missing the required 'name' field.
+Entry: $rendered
+Expected schema: { name: <SERVER NAME>, message: <TEXT>, hint: <TEXT> }
+"@
+            }
+            $i++
+        }
+    }
+
+    # cli_tools: each entry must have 'name'
+    $cliTools = Get-ManifestEntryField -Entry $requires -Field 'cli_tools'
+    if ($cliTools) {
+        $i = 0
+        foreach ($ct in @($cliTools)) {
+            $toolName = Get-ManifestEntryField -Entry $ct -Field 'name'
+            if (-not $toolName) {
+                $rendered = Format-ManifestEntryForError -Entry $ct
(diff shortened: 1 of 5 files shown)


andresharpe (approved):
**pwsh-review** — verdict: ship it. 0 blocker, 0 major, 2 minor, 1 nit.

Static analysis (scoped to changed files): PSScriptAnalyzer 0, InjectionHunter 0, Gitleaks 0. Compatibility pass had 3 hits inside diff hunks, all `PSAvoidUsingWriteHost` in `tests/Test-WorkflowManifest.ps1`, which is excluded by the project PSSA settings (test harness pattern).

Five perspectives — pwsh idioms, security, history: all clean. Two minors and one nit below.

### minor

**[minor] (82) `core/runtime/modules/workflow-manifest.ps1:351,368,385,748`** — five new `throw "string"` sites in a framework module

`standards.md` (Error handling) prefers `Write-Error -ErrorAction Stop -Category <Specific> -ErrorId <Stable>` for production errors and explicitly says `throw "string"` is discouraged in framework modules. Pre-PR `workflow-manifest.ps1` had zero throw sites; all five are new. The pattern is illustrated in `.pwsh-review/patterns/pattern-error-handling.ps1` (e.g. ErrorId `BotRootNotFound`).

Fix: replace each `throw "..."` with `Write-Error -Message "..." -Category InvalidData -ErrorId 'EnvVarsMissingVar' -ErrorAction Stop` (with section-specific IDs: `EnvVarsMissingVar`, `McpServersMissingName`, `CliToolsMissingName`). Update `Test-WorkflowManifest.ps1` to assert on `$_.FullyQualifiedErrorId` instead of message substrings.

**[minor] (82) `core/ui/modules/ProductAPI.psm1:336-338`** — empty `$manifestName` bypasses `<unknown>` default

`Get-PreflightResults` reads `$manifestName` from the manifest and passes it as `-WorkflowName` to `Convert-ManifestRequiresToPreflightChecks`. PowerShell parameter defaults only apply when the parameter is **omitted**, not when an empty value is passed — so a manifest with `name: ""` surfaces an error message that reads `workflow ''` instead of `workflow '<unknown>'`. `Test-WorkflowManifestSchema` already has an explicit fallback for the same case (`workflow-manifest.ps1:250-253`); `Convert-ManifestRequiresToPreflightChecks` does not.

Fix: add the same fallback inside `Convert-ManifestRequiresToPreflightChecks` (`if (-not $WorkflowName) { $WorkflowName = '<unknown>' }`) or guard at the call site.

### nit

**[nit] (55) `core/runtime/modules/workflow-manifest.ps1:223`** — `Test-WorkflowManifestSchema` returns string[]

`Test-*` is conventionally a yes/no question that returns `[bool]` (Test-Path, Test-Connection, and `Test-ValidWorkflowDir` already in this file at line 82). A function that returns a collection of error strings reads more like `Get-WorkflowManifestSchemaError`. Standards don't pin return-type semantics per verb, so this stays a nit; rename if convenient.

### What looked right

- **Diff-bug**: contract change from silent-skip to throw is consistent with the repo direction (PR #318 split intentional skips from framework errors). All call sites of the modified `Convert-ManifestRequiresToPreflightChecks` and `New-EnvLocalScaffold` were checked for the new throw path; `init-project.ps1` and `workflow-add.ps1` both validate before scaffolding and clean up `$wfTargetDir` on failure.
- **Security**: thrown strings are surfaced through `Get-PreflightResults` to the UI, where `escapeHtml` and `textContent` handle the rendering safely. `Format-ManifestEntryForError` builds an inert string; never `iex`'d. `ConvertFrom-Yaml` (YamlDotNet under powershell-yaml) has no Python-style tag deserialization.
- **History**: the original silent-skip was introduced in commit `9f4dbc6` (Mar 2026, #79) without explicit rationale arguing for permissiveness. Issue #319 has only this single commit; no prior partial fix. Test patterns match existing `Test-WorkflowManifest.ps1` style (try/catch flag pattern at lines 875, 1021, 1045).
- **Tests**: 142 new lines covering missing `var`, missing `name`, valid manifest, and multiple-error aggregation.

PR #410, opened 2026-05-09
fix(ui): surface nested skills/agents in Workflows tab (#406)

kabaogluemre:
## Summary

Closes #406 — registry-installed workflows that organise skills/agents in nested folders (e.g. `recipes/skills/overrides/group-1/phase-x/SKILL.md`) had their nested entries hidden in the Workflows tab. The tab listed only top-level folders, including a misleading bare `overrides` entry that has no `SKILL.md` of its own.

This PR fixes both halves of the bug:

**Backend (`/api/workflows/installed`):**
- New `Get-RecipeFolders` helper in `core/runtime/modules/workflow-manifest.ps1` walks `recipes/skills/` (and `recipes/agents/`) recursively, surfacing only folders that directly contain a `SKILL.md` / `AGENT.md` marker. Returns forward-slash relative paths (`overrides/group-1/phase-x`).
- Bare intermediate folders without their own marker are no longer surfaced.
- Recursion is depth-capped at 4 to guard against pathological trees.
- The existing `skills:` array override in `workflow.yaml` continues to short-circuit the walk.

**Frontend (`workflow.js` lazy-load):**
- `loadWfGroupItems` previously deduplicated items by `parts[0]` of the file path, collapsing every nested skill under the same top-level folder (`overrides`) into a single entry. Dedup is now keyed on the full parent-folder path, so each nested skill renders as its own row.
- When a skill folder contains both `SKILL.md` and helpers (e.g. `README.md`), the click target now prefers the marker file.

## Changes

- `core/runtime/modules/workflow-manifest.ps1` — adds `Get-RecipeFolders` (iterative DFS, `-LiteralPath`, depth-capped, marker-aware).
- `core/ui/server.ps1` — `/api/workflows/installed` agents/skills enumeration calls `Get-RecipeFolders` instead of non-recursive `Get-ChildItem`.
- `core/ui/static/modules/workflow.js` — dedup keyed on full parent-folder path; marker-file preference for click target.
- `tests/Test-Components.ps1` — Layer 2 unit tests for `Get-RecipeFolders` covering top-level, nested, intermediate exclusion, missing-marker exclusion, MaxDepth cap, missing base dir.

## Test plan

- [x] Layer 2 unit tests pass (6 new assertions for `Get-RecipeFolders`)
- [x] Manual reproduction in dashboard with 3 flat + 6 nested skills + 1 helper README + 1 empty folder:
  - Skills badge count = 9
  - All 9 skills render with full nested paths (`overrides/group-1/phase-x`, etc.)
  - Bare `overrides` and `empty-folder` are not shown
  - Clicking `overrides/group-1/phase-x` opens `SKILL.md`, not the helper `README.md`
- [x] Default flat-skill workflows render unchanged (no regression)
- [x] Manifest-declared `skills:` array still short-circuits the recursive walk

### DIFF
--- a/core/runtime/modules/workflow-manifest.ps1
+++ b/core/runtime/modules/workflow-manifest.ps1
@@ -138,6 +138,66 @@ function Test-ValidWorkflowDir {
     }
 }
 
+function Get-RecipeFolders {
+    <#
+    .SYNOPSIS
+    Recursively discover recipe folders that contain a given marker file.
+
+    .DESCRIPTION
+    Walks $BaseDir looking for folders that directly contain $MarkerFile
+    (e.g. SKILL.md or AGENT.md). Returns each match as its forward-slash path
+    relative to $BaseDir, so nested folders like
+    `overrides/group-1/phase-x/SKILL.md` surface as `overrides/group-1/phase-x`.
+
+    Intermediate folders without their own marker file are not surfaced — only
+    leaf folders that genuinely contain a recipe show up. Recursion is
+    depth-capped so pathological trees don't impact response time.
+
+    Used by /api/workflows/installed in server.ps1 to expose registry-added
+    nested skills/agents in the Workflows tab. See issue #406.
+    #>
+    param(
+        [Parameter(Mandatory)]
+        [string]$BaseDir,
+
+        [Parameter(Mandatory)]
+        [string]$MarkerFile,
+
+        [int]$MaxDepth = 4
+    )
+
+    if (-not (Test-Path -LiteralPath $BaseDir)) { return @() }
+
+    $results = [System.Collections.Generic.List[string]]::new()
+    $rootFull = (Resolve-Path -LiteralPath $BaseDir).ProviderPath.TrimEnd('\','/')
+
+    $stack = [System.Collections.Generic.Stack[object]]::new()
+    $stack.Push(@{ Path = $rootFull; Depth = 0 })
+
+    while ($stack.Count -gt 0) {
+        $frame = $stack.Pop()
+        $current = $frame.Path
+        $depth   = $frame.Depth
+
+        if ($depth -gt 0) {
+            $marker = Join-Path $current $MarkerFile
+            if (Test-Path -LiteralPath $marker -PathType Leaf) {
+                $rel = $current.Substring($rootFull.Length).TrimStart('\','/') -replace '\\','/'
+                if ($rel) { $results.Add($rel) }
+            }
+        }
+
+        if ($depth -ge $MaxDepth) { continue }
+
+        $children = Get-ChildItem -LiteralPath $current -Directory -ErrorAction SilentlyContinue
+        foreach ($child in $children) {
+            $stack.Push(@{ Path = $child.FullName; Depth = $depth + 1 })
+        }
+    }
+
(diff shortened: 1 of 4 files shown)


andresharpe (approved):
@
**pwsh-review** — verdict: ship it. 0 blocker, 0 major, 2 minor, 1 nit.

Static analysis (scoped to changed files): PSScriptAnalyzer 0, compatibility 0, InjectionHunter 5 (all in pre-existing server.ps1 code outside the diff hunk at 1931-1944), Gitleaks 0.

Five perspectives — pwsh idioms, diff bugs, security, history: all clean. Conventions surfaced two minors and one nit, listed below; non-blocking, but worth a follow-up if convenient.

### minor

**[minor] (85) `core/runtime/modules/workflow-manifest.ps1:141`** — `Get-RecipeFolders` uses a plural noun

`standards.md` says "Cmdlet nouns are singular." Sibling functions in this file (`Read-WorkflowManifest`, `Test-ValidWorkflowDir`, `Get-ActiveWorkflowManifest`) all follow that rule, as do PowerShell built-ins (`Get-ChildItem`, `Get-Process`).

Fix: rename to `Get-RecipeFolder` and update the two call sites in `core/ui/server.ps1` and the assertions in `tests/Test-Components.ps1`.

**[minor] (85) `core/runtime/modules/workflow-manifest.ps1:192`** — `-ErrorAction SilentlyContinue` without an inline comment

```powershell
$children = Get-ChildItem -LiteralPath $current -Directory -ErrorAction SilentlyContinue
```

`standards.md` requires an inline comment justifying every `-ErrorAction SilentlyContinue`. The intent is obvious (skip unreadable subfolders during traversal), but the rule is explicit.

Fix: add `# Skip unreadable subfolders during traversal — they cannot contain a recipe marker we can use.`

### nit

**[nit] (70) `core/runtime/modules/workflow-manifest.ps1:142-158`** — incomplete comment-based help

CBH has `.SYNOPSIS` and `.DESCRIPTION` but no `.PARAMETER` entries for `BaseDir` / `MarkerFile` / `MaxDepth` and no `.EXAMPLE`. Internal helper, so this is a nit; the function will gain external callers if registry-added recipes catch on, in which case `.PARAMETER` lines would help.

### What looked right

- **Diff-bug**: the contract change for `agents` / `skills` arrays (now path-bearing identifiers like `overrides/group-1/phase-x` instead of bare folder names) is safe — only `.length` is consumed in workflow.js, and the studio UI uses separate `agentFiles` / `skillFiles` fields.
- **Security**: `Get-RecipeFolders` walks directories with a depth cap of 4 and reads no file contents (only `Test-Path -PathType Leaf`). `BaseDir` flow is `Join-Path $wfDir "recipes\agents"` — never sourced from HTTP input. JSON output is ConvertTo-Json escaped; no XSS surface introduced.
- **History**: the previous one-deep enumeration in `server.ps1:1931` (commit `1c5fba0`) was not a deliberate non-recursive choice — the commit message was about returning correct folder names. The JS dedup-by-display-name (commit `9f4dbc6`) was about collapsing helper files into one entry, not about flattening nesting. Both intents are preserved.
- **Tests**: cover the happy path, the depth cap, the marker filter, and intermediate-folder exclusion.
@

PR #432, opened 2026-05-20
fix: gitignore dotbot framework files; treat them as installed deps not committed state

bboyukbas:
## Summary

Closes #431.

Treats dotbot framework files like installed dependencies — they live on each developer's disk after `dotbot init`, gitignored in the consumer repo. Workspace content (`.bot/workspace/`) stays tracked.

## What changes

| File | Lines | What |
|---|---|---|
| `core/mcp/modules/FrameworkIntegrity.psm1` | +51 / -48 | Tolerate gitignored framework files. SHA256 manifest does all tamper detection; `git status` step runs only when `.bot/` is tracked. |
| `scripts/init-project.ps1` | +131 / -44 | Add framework paths to consumer `.gitignore`. One-time `git rm --cached` migration for projects that previously committed framework files. Initial / first-init commits stage only `.bot/workspace/` + `.mcp.json` + `.gitignore`. `--force` is local-only (no commit). Invert post-init guard (workspace MUST be tracked; framework SHOULD be gitignored). |
| `core/hooks/scripts/commit-bot-state.ps1` | +19 / -9 | Stage `.bot/workspace/` only, never framework or runtime. |

Net: 3 files, +201 / -101. No new files, no deletions.

## Effect

- **New installs:** initial commit shrinks from ~70k lines (~228 framework files) to a few `.gitkeep`s plus `.mcp.json` and the `.gitignore` updates.
- **Existing installs:** next `dotbot init` self-heals — single `chore: untrack dotbot framework files` commit untracks the previously-committed framework, and from then on the bloat stops.
- **Autonomous task saves:** small commits (workspace deltas only) instead of multi-k chore commits.
- **`dotbot init --force`:** no longer creates a commit; framework refresh is purely local.

## Migration semantics

The migration block runs on every `dotbot init` invocation, but only does work when both:
1. The repo already has commits (`git rev-parse HEAD` succeeds), AND
2. At least one framework path is currently tracked (`git ls-files <path>` returns non-empty)

If neither condition holds (fresh repo or already-migrated install), the migration is a silent no-op.

The migration commit is bypass-the-pre-commit-guard via `DOTBOT_FORCE_COMMIT=1` (same pattern as the existing initial-commit and `--force` flows).

## Test plan

- [x] All three files are valid PowerShell — no parse errors.
- [x] `Test-Structure.ps1` assertions still hold: `Test-FrameworkTracked`, `Test-FrameworkIntegrity`, `Invoke-FrameworkIntegrityGate`, `Test-DotbotManifest`, `git check-ignore`, and `missing-manifest` are all still present in `FrameworkIntegrity.psm1` (verified by file-content grep).
- [ ] Manual: `dotbot init` on a fresh repo → initial commit is small (workspace `.gitkeep`s + config only).
- [ ] Manual: `dotbot init` on a repo with previously-committed framework → migration commit untracks framework files; subsequent `git log` shows it as its own chore commit.
- [ ] Manual: autonomous-task start triggers `commit-bot-state.ps1` → commit message lists only `.bot/workspace/*` paths.
- [ ] Manual: `dotbot init --force` on a Plan-D-converted project → no commit, framework refreshes locally, working tree stays clean.
- [ ] Manual: `Test-FrameworkIntegrity` on a project with gitignored framework + intact `.bot/.manifest.json` → success, not the legacy `gitignored = failure` response.
- [ ] Manual: edit `.bot/core/foo.psm1` directly → `Test-FrameworkIntegrity` flags it via the manifest hash mismatch (proves manifest-only detection still works).

## Notes for reviewers

The most controversial assertion is **"framework files should be gitignored, not committed."** This contradicts the previous load-bearing comment in `FrameworkIntegrity.psm1`:

> The integrity model assumes .bot/ is committed to git in the target project. That is how dotbot init bootstraps the framework and is a load-bearing assumption.

My rationale for changing the premise:

1. The SHA256 manifest catches every form of tampering regardless of git tracking. The "git status --porcelain" backup was duplicative — it only added value for users who edited files but didn't run `dotbot init` to regenerate the manifest. The manifest path covers that case directly: a hand-edit changes the SHA, the verify hook trips.
2. Framework files are functionally indistinguishable from `node_modules/`: installed by an init tool, version-pinned via `~/dotbot/version.json`, and a tampered-with copy can always be repaired by `dotbot init --force`.
3. Worktree state sharing uses Windows junctions / Unix symlinks (per `WorktreeManager.psm1`). These live on disk and are independent of git tracking.
4. The pre-commit framework-file-protection hook installed by `init-project.ps1` becomes mostly redundant for gitignored files (gitignored files can't be staged without `-f`). It remains useful as a safety net for `git add -f` accidents and is unchanged in this PR.

If this premise is wrong, I'd appreciate a concrete failure mode: a scenario where framework files are gitignored, `.bot/.manifest.json` is intact, but tampering goes undetected. I couldn't construct one.

## Why not the alternatives

- **Plan A (gitignore all of `.bot/`):** breaks team-shared workspace content (tasks, decisions, plans, product docs).
- **Plan B (only fix `commit-bot-state.ps1`):** stops the recurring autonomous-save bloat but leaves the +70k from `dotbot init` itself — every fresh dotbot install still bloats the consumer.
- **Plan C (framework lives only in `~/dotbot/`, no copies into `.bot/` at all):** architecturally cleanest but requires refactoring every script path-resolution, the MCP config in `.mcp.json`, and the worktree-junction setup. Probably a 3-5 PR refactor. This PR (Plan D) is the practical sweet spot.

## Test environment

- dotbot `v3.5`
- Windows 11, PowerShell 7.5, Git 2.x
- Branch: `bboyukbas/dotbot:chore/no-framework-bloat`
- Diagnosis came from real-world bloat on a large .NET Framework consumer repo (~90 projects) with multiple Claude Code worktrees.


carlospedreira:
Thanks for the PR, @bboyukbas. Going to close this.

The framework-in-`.bot/` approach was a deliberate design choice enabling straightforward integrity verification via co-located SHA256 manifests and git status. It is not bloat.

That said, the concern you raised is valid. v4 addresses it architecturally by eliminating the framework copy into `.bot/` entirely — framework files are resolved from `DOTBOT_HOME` at runtime, with `.bot/` holding only user workspace data. Framework co-location will remain available as an opt-in for environments that need it, but the default becomes the lightweight model you were aiming for.

PR #434, opened 2026-05-20
fix: skip rate_limit_event warning when API status is allowed

bbolek-ap:
## Summary

Fixes #433

When the Claude API emits a `rate_limit_event` with `status: "allowed"` (an informational event confirming the request was **not** blocked), dotbot incorrectly fired a `⚠ RATE LIMIT` warning. This is a false positive.

**Root cause — two issues:**

1. The regex `error.*?rate_limit` matches any JSON line containing `rate_limit` anywhere in the string (e.g. `"type":"rate_limit_event"`), not just actual error responses.
2. Neither file checked `rate_limit_info.status` before emitting the warning.

## Changes

Added an early return in both files after JSON parsing — if the event is a `rate_limit_event` with `status: "allowed"`, it's treated as informational and skipped:

```powershell
if ($jsonObj.type -eq "rate_limit_event" -and $jsonObj.rate_limit_info.status -eq "allowed") {
    return  # informational only — request was allowed, not blocked
}
```

## Files Changed

| File | Change |
|------|--------|
| `core/runtime/ClaudeCLI/ClaudeCLI.psm1` | Added early return after `ConvertFrom-Json` (line ~703) |
| `core/runtime/ProviderCLI/parsers/Parse-ClaudeStream.ps1` | Added early return after `ConvertFrom-Json` (line ~34) |

## Test

Trigger a `rate_limit_event` with `status: "allowed"` from the API — no `⚠ RATE LIMIT` warning should appear. An event with `status: "blocked"` or `$jsonObj.error -eq "rate_limit"` should still fire the warning as before.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/core/runtime/ClaudeCLI/ClaudeCLI.psm1
+++ b/core/runtime/ClaudeCLI/ClaudeCLI.psm1
@@ -698,7 +698,7 @@ function Invoke-ClaudeStream {
             if ($line.Length -eq 0) { return }
             
             # Check for rate limit in JSON responses (#391: includes org/monthly quota wording)
-            if ($line[0] -eq '{' -and $line -match "hit your.*?limit|out of extra usage|error.*?rate_limit") {
+            if ($line[0] -eq '{' -and $line -match "hit your.*?limit|out of extra usage|error.*?rate_limit|rate_limit_event") {
                 try {
                     $jsonObj = $line | ConvertFrom-Json -ErrorAction Stop
                     # Skip informational rate_limit_event when request was allowed (#433)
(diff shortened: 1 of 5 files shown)


carlospedreira:
Closing — this targets `core/runtime/ClaudeCLI/ClaudeCLI.psm1` and `core/runtime/ProviderCLI/parsers/Parse-ClaudeStream.ps1`, both of which are removed in the imminent v4 rewrite (the PowerShell stream-parser layer no longer hand-parses `rate_limit_event`, so this false positive cannot recur). Given v4 is shipping shortly and the warning is cosmetic, not worth patching the current parser. Thanks @bbolek-ap.

PR #446, opened 2026-05-25
fix(init): preserve other workflows on -Force re-init + unbreak --Force in workflow dispatcher (#442)

kabaogluemre:
## Summary

`dotbot init -Force -Workflow X` was silently wiping every other installed
workflow:

- **On disk:** the legacy-content migration recursed into `.bot/workflows/<name>/`
  and deleted `workflow.yaml` + `recipes/{prompts,includes,research}` in every
  sibling workflow — those paths are canonical post-PR-5 content, not legacy.
- **In settings:** `.bot/settings/` was wiped and re-copied from the template,
  then `installed_workflows` was overwritten with just the workflow installed
  this run. No read-merge-write.

The issue's documented workaround (`dotbot workflow add <name> --Force`) also
didn't work: the `Invoke-Workflow` dispatcher passed switches via array splat,
so PowerShell errored with "A positional parameter cannot be found that
accepts argument '--Force'".

Closes #442.

### `scripts/init-project.ps1`

1. **Drop the recursive migration into `.bot/workflows/<name>/`.** Root-level
   pre-PR-5 cleanup (`.bot/workflow.yaml`, `.bot/recipes/{prompts,includes,research}/`)
   is kept; only the recursive call that treated post-PR-5 workflow content as
   legacy is removed.
2. **Preserve `installed_workflows` across `-Force`.** Capture the existing
   list before settings/ is wiped (mirrors the existing `instance_id`
   preservation), then write the union of (preserved, this run's `-Workflow`)
   deduped and validated against `Test-ValidWorkflowDir`. Lifted the write
   site out of the `if ($Workflow)` block so re-init **without** `-Workflow`
   also preserves the list.

### `scripts/install-global.ps1` (CLI heredoc → `bin/dotbot.ps1`)

3. **Fix `Invoke-Workflow` dispatcher.** Use the same hashtable-splat parsing
   pattern that `Invoke-Registry` already uses, so `--Force` and `-Force` bind
   as named switches. Positional workflow name is mapped to `-Name`. Required
   `[array]` cast on `$wfRest` to avoid the PowerShell single-element-array
   unwrap that turns `'start-from-prompt'` into the string `'s'` after `[0]`
   indexing — same regression class as the existing `@wfExtra` null test.

### `tests/Test-WorkflowIntegration.ps1`

4. **18 new regression tests** under two sections:
   - `INIT -FORCE MULTI-WORKFLOW SAFETY (#442)` — covers all acceptance
     criteria: pre-existing workflow files survive, `installed_workflows` is
     unioned with no duplicates, repeat `init -Force` is idempotent, fresh
     `init -Force -Workflow X` still produces `[X]`, root-level legacy
     migration still runs.
   - `CLI DISPATCHER --Force NORMALIZATION (#442)` — `dotbot workflow add X
     --Force` and `dotbot workflow add X -Force` both dispatch without
     "positional parameter cannot be found" errors.

### DIFF
--- a/scripts/init-project.ps1
+++ b/scripts/init-project.ps1
@@ -217,19 +217,16 @@ function Invoke-BotFolderMigration {
     # had .bot/workflow.yaml plus old prompts/includes/research at .bot/recipes/.
     # None of these have a consumer in the new layout —
     # remove them so a re-init does not leave a confusing hybrid tree.
+    #
+    # Scope: ROOT ONLY. Do NOT recurse into .bot/workflows/<name>/ — in the
+    # post-PR-5 layout those files (workflow.yaml, recipes/prompts/, etc.) are
+    # the canonical content of each installed workflow, not legacy residue.
+    # See issue #442.
     $legacyManifest = Join-Path $Dir "workflow.yaml"
     if (Test-Path $legacyManifest) { Remove-Item -Path $legacyManifest -Force }
     foreach ($legacy in @("recipes/prompts", "recipes/includes", "recipes/research")) {
         $legacyPath = Join-Path $Dir $legacy
         if (Test-Path $legacyPath) { Remove-Item -Path $legacyPath -Recurse -Force }
-    }
-
-    # Migrate installed workflow subdirectories
-    $wfDir = Join-Path $Dir "workflows"
-    if (Test-Path $wfDir) {
-        Get-ChildItem $wfDir -Directory | ForEach-Object {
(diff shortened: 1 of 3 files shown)


carlospedreira:
@kabaogluemre Closing this. The underlying issue (#442) is resolved in v4: the v4 init/workflow layout drops the recursive workflow deletion and `installed_workflows` behavior this PR fixed for main, and the `workflow add --Force` dispatcher fix is already on v4 (commit 2191b25). Not merging to main as it is superseded by v4.

PR #499, opened 2026-06-18
fix(ui): use escapeAttr for attribute-context interpolation in action…

kabaogluemre:
## Summary

`src/ui/static/modules/actions.js` interpolated dynamic values into HTML **attribute** contexts using `escapeHtml()`. `escapeHtml()` escapes `&`, `<`, `>` but **not** `"` or `'`, so a value containing a quote can break out of the attribute and inject a new attribute or HTML structure.

This sweep converts every attribute-context interpolation to `escapeAttr()` (which also escapes `"` and `'`), reserving `escapeHtml()` for text-node interpolation. No behavior change for canonical-format values.

Closes #477

## Changes

Swapped `escapeHtml` → `escapeAttr` at **15** attribute-context interpolation points in `src/ui/static/modules/actions.js`:

| Region | Lines |
|---|---|
| Approval card (`data-task-id`, `data-key`, `data-label`) | 306, 350, 351, 368 |
| Split decision (`data-task-id`) | 514 |
| Workflow-launch interview (`data-process-id`, `data-question-id`, `data-key`, `data-label`, `data-question-key`) | 552, 560, 566, 567, 568, 580, 582 |
| File-remove buttons (`data-task-id`, `data-key`) | 996, 1045 |
| Review card (`data-task-id`) | 1357 |

The original issue cited 5 `data-task-id` spots; this PR extends to the full sweep (the issue title and "what you expected" call for *every* attribute-context value to use `escapeAttr`). Text-node interpolations (`answer-key`, `answer-label`, file names, question text, etc.) intentionally remain on `escapeHtml`.

## Why

`escapeAttr()` was already in scope and used correctly elsewhere in the file. The convention should hold uniformly so a future field swap (e.g. a user-supplied name reaching one of these attributes) does not silently introduce an injection sink.

### DIFF
--- a/src/ui/static/modules/actions.js
+++ b/src/ui/static/modules/actions.js
@@ -303,7 +303,7 @@ function renderQuestionItem(item) {
 
     if (questionType === 'approval') {
         return `
-        <div class="action-item" data-task-id="${escapeHtml(item.task_id)}" data-type="question" data-question-type="approval">
+        <div class="action-item" data-task-id="${escapeAttr(item.task_id)}" data-type="question" data-question-type="approval">
             <div class="action-item-header">
                 <span class="action-item-type question">Approval</span>
                 <span class="action-item-task">${escapeHtml(item.task_name)}</span>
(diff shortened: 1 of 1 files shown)


carlospedreira (approved):
Reviewed the attribute-context escaping sweep. Using escapeAttr for data-* values and keeping escapeHtml for text nodes is the right fix.

PR #501, opened 2026-06-18
fix: align workflow task progress counts between Workflow and Overvie…

kabaogluemre:
## Summary

The Workflow tab and Overview tab showed different task-progress numbers for the same workflow. Issue #453 reported the Workflow page showing fewer completed tasks than the Overview page.

**Root cause:** the two surfaces are fed by different data sources with different counting rules:

- **Overview** (`renderOverviewWorkflowPhases`) and the Workflow-tab phase widget (`renderWorkflowTaskProgress`) read from `/api/state` and count `(done + skipped) / total`.
- The **Workflow nav tree** (`renderWorkflowDetailPanel`) and meta sidebar read from `/api/workflows/installed`, which counted `done / total` — `skipped` tasks were included in `total` but **not** in `done`.

Result: any workflow with skipped tasks displayed a lower percentage / count in the Workflow nav tree than on Overview.

## Changes

**Backend — `src/ui/server.ps1`** (`/api/workflows/installed`)
- Track `skipped` in the per-workflow task bucket and increment it in the status switch.
- Add `skipped = 0` to the two fallback buckets for shape consistency.

**Frontend — `src/ui/static/modules/workflow.js`**
- Nav-tree progress bar now computes `done = (tasks.done || 0) + (tasks.skipped || 0)`.
- Change-detection cache key now includes `skipped`, so a `todo → skipped` transition triggers a re-render (otherwise the bar would show a stale value).
- Meta sidebar "Tasks" row now shows `(done + skipped) / total`.

All workflow-progress surfaces now use the same `(done + skipped) / total` formula.

Closes #453

### DIFF
--- a/src/ui/server.ps1
+++ b/src/ui/server.ps1
@@ -2236,13 +2236,14 @@ $docContext
                             $wf = Get-CachedTaskWorkflow -File $_
                             $key = if ($wf) { $wf } else { '__default__' }
                             if (-not $tasksByWorkflow.ContainsKey($key)) {
-                                $tasksByWorkflow[$key] = @{ todo = 0; in_progress = 0; done = 0; total = 0 }
+                                $tasksByWorkflow[$key] = @{ todo = 0; in_progress = 0; done = 0; skipped = 0; total = 0 }
                             }
                             $tasksByWorkflow[$key]['total']++
                             switch ($statusDir) {
                                 'todo'        { $tasksByWorkflow[$key]['todo']++ }
                                 'in-progress' { $tasksByWorkflow[$key]['in_progress']++ }
                                 'done'        { $tasksByWorkflow[$key]['done']++ }
+                                'skipped'     { $tasksByWorkflow[$key]['skipped']++ }
                             }
                         }
                     }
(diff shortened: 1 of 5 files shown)


carlospedreira:
@kabaogluemre This fix makes sense to me.

One small follow-up idea: the meaning of “completed” is now effectively `done + skipped`, but that calculation still lives in a few different render paths. For this PR, the targeted fix is probably fine. But if we want to avoid this drifting again, a small frontend helper could make it clearer, for example something like `getCompletedTaskCount(counts)` / `getTaskProgress(counts)` that returns `done + skipped`, `total`, and the percentage.

That helper could then be used by the Workflow nav, the Workflow metadata row, and the Overview workflow progress rendering. I would not push for a larger API/state refactor here; a small shared frontend helper is probably enough.

kabaogluemre:
> @kabaogluemre This fix makes sense to me.
> 
> One small follow-up idea: the meaning of “completed” is now effectively `done + skipped`, but that calculation still lives in a few different render paths. For this PR, the targeted fix is probably fine. But if we want to avoid this drifting again, a small frontend helper could make it clearer, for example something like `getCompletedTaskCount(counts)` / `getTaskProgress(counts)` that returns `done + skipped`, `total`, and the percentage.
> 
> That helper could then be used by the Workflow nav, the Workflow metadata row, and the Overview workflow progress rendering. I would not push for a larger API/state refactor here; a small shared frontend helper is probably enough.

@carlospedreira thank you for your review. I have centralized the calculation as you suggested and committed the change.

carlospedreira (approved):
Looks good. The follow-up helper centralizes the completed-task calculation and the checked progress surfaces now use it.

PR #502, opened 2026-06-18
fix(ui): show all completed tasks on Roadmap via a "Show more" contro…

kabaogluemre:
## Summary

The **Roadmap** tab's *Done* column showed an inaccurate set of tasks: the count badge reported the true total (e.g. `15`) while only the first **10** were rendered, with no visible way to reach the rest. Fixes #454.

## Root cause

`updatePipelineColumn` caps rendering at `pipelineDisplayLimits[containerId]` (default **10**) and relies solely on a `scroll` listener to load more (`+5` per scroll-to-bottom). That listener is bound to the **column's own** scroll region, so the remaining tasks become unreachable whenever:

- the user scrolls the page instead of the column's internal scrollbar (most common),
- the column doesn't overflow, so no `scroll` event ever fires (reproducible at lower resolutions / shorter columns),
- nothing signals that more tasks exist below the fold.

The result: count says 15, list shows 10, and the gap reads as "missing tasks".

## Fix

Add a visible **"Show N more"** affordance at the bottom of any truncated column:

- `N` is dynamic (`total − visible`) — e.g. *Show 5 more* at 15 tasks, *Show 10 more* at 20.
- One click reveals all remaining tasks, guaranteeing the rendered list matches the count badge.
- Coexists with the existing scroll-based loader; on a scrollable column the autoload still works, but the button removes the dependency on a hidden scroll gesture.
- Styled with existing design tokens (CRT/retro), consistent with the column.

Source list is already capped at 100 server-side, so the rendered set stays bounded.

## Files changed

- `src/ui/static/modules/ui-updates.js` — render the load-more button + click handler
- `src/ui/static/css/views.css` — `.pipeline-load-more` styling

## Testing

- Simulated 15 and 20 completed tasks → badge `15`/`20`, column renders 10 + *Show 5 more* / *Show 10 more*; click reveals the full set and the button disappears.
- Reproduced the original bug by lowering the window resolution (column stops overflowing, scroll-autoload never fires) and confirmed the button resolves it.
- Verified Overview tab totals stay consistent with the Roadmap column.

### DIFF
--- a/src/ui/modules/StateBuilder.psm1
+++ b/src/ui/modules/StateBuilder.psm1
@@ -388,7 +388,9 @@ function Get-BotState {
     # Per-workflow task counts accumulator
     $workflowCounts = @{}
 
-    # Get recent completed tasks (last 100 for infinite scroll)
+    # Get recent completed tasks. The Roadmap "Done" column is a recent-history
+    # view, not an exhaustive archive — cap to the newest 20 to keep the state
+    # payload and the rendered column bounded (PR #502 review).
     $recentCompleted = @()
     if ($doneTasks.Count -gt 0) {
         $recentCompleted = $doneTasks |
@@ -435,7 +437,7 @@ function Get-BotState {
                 }
             } | Where-Object { $_ -ne $null } |
             Sort-Object { if ($_.completed_at) { [DateTime]$_.completed_at } else { [DateTime]::MinValue } } -Descending |
-            Select-Object -First 100
+            Select-Object -First 20
     }
 
     # Get skipped tasks list
@@ -473,7 +475,9 @@ function Get-BotState {
                 } catch {
                     $null
                 }
-            } | Where-Object { $_ -ne $null }
+            } | Where-Object { $_ -ne $null } |
+            Sort-Object { if ($_.updated_at) { [DateTime]$_.updated_at } else { [DateTime]::MinValue } } -Descending |
+            Select-Object -First 10
     }
 
     # Get analysing tasks list
--- a/src/ui/static/css/views.css
+++ b/src/ui/static/css/views.css
@@ -210,6 +210,29 @@
     background: var(--color-secondary);
 }
 
+.pipeline-load-more {
+    display: block;
+    width: 100%;
+    margin-top: 4px;
+    padding: 6px 12px;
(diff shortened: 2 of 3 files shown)


carlospedreira:
@kabaogluemre This makes the hidden Done-column items reachable from the UI, which is good.

One behavior to confirm: the list is still limited by the StateBuilder payload. `src/ui/modules/StateBuilder.psm1` currently caps `recent_completed` to the newest 100 done tasks with `Select-Object -First 100`, while `skipped_list` is sent without the same cap. The frontend then combines those in `src/ui/static/modules/ui-updates.js` as `recent_completed + skipped_list`.

I think a cap makes sense to keep the payload and Roadmap bounded, but we should probably cap skipped tasks too. We may also want a smaller cap here, for example the last 20 done tasks and last 10 skipped tasks, since this view only needs enough recent history to be useful.

kabaogluemre:
> @kabaogluemre This makes the hidden Done-column items reachable from the UI, which is good.
> 
> One behavior to confirm: the list is still limited by the StateBuilder payload. `src/ui/modules/StateBuilder.psm1` currently caps `recent_completed` to the newest 100 done tasks with `Select-Object -First 100`, while `skipped_list` is sent without the same cap. The frontend then combines those in `src/ui/static/modules/ui-updates.js` as `recent_completed + skipped_list`.
> 
> I think a cap makes sense to keep the payload and Roadmap bounded, but we should probably cap skipped tasks too. We may also want a smaller cap here, for example the last 20 done tasks and last 10 skipped tasks, since this view only needs enough recent history to be useful.

@carlospedreira , thanks for your review. I have limited and ordered skipped tasks and lowered the limitation to show recent 20 done and 10 skipped tasks. But just fyi, after more than 20 done tasks, the counter is going to be inconsistent what shown in list.

carlospedreira:
@kabaogluemre thanks, that makes sense. I think this is part of a broader frontend cleanup we still need, rather than something to keep fixing field by field.

The Done column is currently mixing total count, capped returned items, and skipped tasks. So for this PR, I’m okay with the bounded list behavior, but I don’t think we should keep layering small UI semantics onto this view until we have that target defined.

carlospedreira (approved):

Issue #356, opened 2026-04-27
MCP server resolves DotbotProjectRoot to the worktree, silently discarding all agent-driven task-state transitions


andresharpe:
## TL;DR

The dotbot MCP server (`core/mcp/dotbot-mcp.ps1`) resolves `$global:DotbotProjectRoot` by walking up from `$PSScriptRoot` until it finds something at `.git`. The `Test-Path` check matches **both** a `.git` directory and the `.git` gitfile that lives at the root of every git worktree. So when the MCP server is launched from a per-task worktree (the standard dotbot convention) it sets `DotbotProjectRoot` to the **worktree root** rather than the main repo. Every MCP tool reads that variable to locate `.bot/workspace/tasks/`, so every agent-driven task-state transition (`task_mark_needs_input`, `task_mark_done`, `task_mark_skipped`, `task_mark_analysed`, `task_answer_question`, etc.) writes to the worktree. `Complete-TaskWorktree` then runs `git checkout -- .bot/workspace/tasks/` on the worktree before the squash-merge, **silently discarding** every one of those writes.

Net effect: any change the agent makes to task state from within a worktree is invisible to the parent. PR #352's loop-escalation fix (move failed execution to `needs-input`) is no-op'd in worktree runs. So is `mark_done`, so is everything else.

## Repro

A live SlashOps dogfood run on `2026-04-27` (process `proc-8a41a3`, started 13:07Z, runner stopped by the operator around 14:01Z when the task hit a privacy-scan blocker). All paths and timestamps below come from `.bot/.control/activity.jsonl` and `.bot/.control/processes/proc-8a41a3.activity.jsonl` of that run.

**Activity log timeline for task `Product Documents` (id `37dcd1a5-9c6f-4d42-b01c-cee834589f12`):**

```
13:30:31  mcp__dotbot__task_mark_needs_input   (called from worktree's Claude session)
13:31:03  Squash-merged to main: Product Documents
13:31:03  Task completed (analyse+execute): Product Documents
```

**State on disk after the merge:**

```
.bot/workspace/tasks/in-progress/product-documents-37dcd1a5.json
  status:           in-progress
  updated_at:       2026-04-27T13:11:27Z   ← original execute-phase timestamp, never updated
  pending_question: null
  completed_at:     null
```

[...]

PR #527, opened 2026-06-22
fix(runtime): use main project root for MCP preflight on task retry (#515)

kabaogluemre:
## Summary

Fixes #515 — `Test-DotbotMcpReadiness` crashing with *"MCP server exited or returned an empty response"* when a task is retried after a failed squash-merge.

## Root cause

The standalone preflight MCP process had `DOTBOT_PROJECT_ROOT` set to the **worktree path**. On task retry the worktree's `.control` junction can be stale — teardown/re-creation is not atomic — so the MCP server cannot resolve `runtime.json` (via `<worktree>/.bot/.control/`) and exits **before the handshake even begins**.

Two existing fixes don't cover this case:

- **#479** retries the MCP *handshake* — but here the process dies pre-handshake, so retrying has no effect.
- **#356** fixed `Resolve-DotbotProjectRoot` via `git rev-parse --git-common-dir` — but that detection is **bypassed** whenever `DOTBOT_PROJECT_ROOT` is set, because `Resolve-ProjectRoot.ps1` returns the env var value verbatim and skips git-based detection (see `Resolve-ProjectRoot.ps1:30-49`).

## Fix

Add an optional `-ProjectRoot` parameter to `Test-DotbotMcpReadiness`. When supplied, it is used for `DOTBOT_PROJECT_ROOT` instead of the worktree path:

```powershell
$psi.Environment['DOTBOT_PROJECT_ROOT'] = if ($ProjectRoot) { $ProjectRoot } else { $WorktreePath }
```

The call site now passes the main project root, which always has a stable `.control/` directory:

```powershell
$mcpReady = Test-DotbotMcpReadiness -WorktreePath $worktreePath -ProjectRoot $projectRoot
```

`WorkingDirectory` stays the worktree (work happens there); only task-state resolution is redirected to the main root — consistent with #356's intent. Backward compatible: callers that omit `-ProjectRoot` fall back to the previous worktree behaviour.

Closes #515


carlospedreira:
@kabaogluemre this highlights a bigger concern with the root model around MCP. Right now, `DOTBOT_PROJECT_ROOT` seems to carry two meanings: the worktree where the agent should operate, and the stable project root where dotbot runtime/task state lives. This PR fixes the preflight path by pointing it at the stable root, but the actual provider MCP config and launch paths still point the MCP server at the worktree, so preflight and the real session can diverge.

I think we should make that split explicit instead of patching one caller. The agent cwd should stay as the worktree, but MCP state/runtime resolution should use a dedicated `DOTBOT_STATE_ROOT`. That would avoid relying on worktree `.control` links being valid during retry/teardown windows, and it would give us tests that cover the actual failure mode rather than only checking the preflight assignment.

kabaogluemre:
> @kabaogluemre this highlights a bigger concern with the root model around MCP. Right now, `DOTBOT_PROJECT_ROOT` seems to carry two meanings: the worktree where the agent should operate, and the stable project root where dotbot runtime/task state lives. This PR fixes the preflight path by pointing it at the stable root, but the actual provider MCP config and launch paths still point the MCP server at the worktree, so preflight and the real session can diverge.
> 
> I think we should make that split explicit instead of patching one caller. The agent cwd should stay as the worktree, but MCP state/runtime resolution should use a dedicated `DOTBOT_STATE_ROOT`. That would avoid relying on worktree `.control` links being valid during retry/teardown windows, and it would give us tests that cover the actual failure mode rather than only checking the preflight assignment.

@carlospedreira , thanks for your review, a dedicated `DOTBOT_STATE_ROOT` is the cleaner approach. Applied it across the resolver, all provider adapters, and the generated MCP configs, so preflight and the real session now share the same model. Also added tests for the actual resolver failure mode.

carlospedreira (approved):

PR #528, opened 2026-06-22
fix(task-input): make interview-answers.json a main-checkout accumulator to stop parallel squash-merge conflicts (#516)

kabaogluemre:
## Summary

Fixes #516 — parallel tasks answering Approve/Reject decisions caused unresolvable
squash-merge conflicts on `.bot/workspace/product/interview-answers.json`.

This file is a **run-level accumulator**: every decision answer is appended to one
shared JSON array. It was being routed to each task's **per-worktree** copy, so
concurrent task branches diverged on the same array and collided at squash-merge,
failing with `rebase_conflict` and escalating the task to `needs-input`.

The fix moves ownership of the accumulator to the **main project checkout** (where
the Product UI already reads from), so concurrent task completions no longer race on
the same file.

## Changes

**1. Route writes to the main checkout** — `Dotbot.TaskInput.psm1`
- `Get-TaskInputProductDir` no longer resolves a per-task worktree path; it always
  returns the main checkout's `workspace/product`.
- Because all parallel tasks now write to the *same* file, `Write-TaskInputInterviewAnswer`
  serialises appends with a cross-process lock (`Invoke-WithTaskLock`, constant key
  `interview-answers`) to prevent read-modify-write lost updates. The answer also
  becomes visible in the Product tab the moment it's recorded, instead of after merge.

**2. Exclude the accumulator from worktree git tracking** — `Dotbot.Worktree.psm1`
- `Ensure-DotbotWorktreeExcludes` adds `.bot/workspace/product/interview-answers.json`
  so git never tracks a worktree-local copy — it can't enter a task branch's
  squash-merge patch at all. Other `product/` docs still flow normally.

**3. Auto-resolve safety net for legacy branches** — `Dotbot.Worktree.psm1`
- `Apply-TaskBranchPatch` now auto-merges an unresolved `interview-answers.json`
  conflict instead of escalating: `Merge-DotbotInterviewAnswers` unions both sides
  by `question_id` (latest `answered_at` wins, deterministic ordering) and stages the
  result. Covers worktrees/branches created before the exclude existed. Any other
  conflict stays unresolved.

## Tests

- `Test-Components.ps1`
  - End-to-end: `Apply-TaskBranchPatch` auto-merges two task branches' answers without
    a `rebase_conflict`; merged file contains both `question_id`s.
  - `Ensure-DotbotWorktreeExcludes` pins `interview-answers.json` into `.git/info/exclude`
    while leaving the rest of `product/` tracked.
- `Test-TaskActions.ps1`
  - Batch answer now lands in the **main checkout** and is **absent from the worktree**.
  - Shared accumulator assertions use containment (not array position), since answers
    from all tasks in the run coexist.

Closes #516


carlospedreira:
@kabaogluemre thanks for working on this. I’m not rejecting the PR, but I don’t think it fixes the right problem.

The issue is shared mutable state across parallel tasks. If multiple tasks write to one file, that file needs to be runtime-owned with a clear concurrency model. I don’t think `.bot/workspace/product/interview-answers.json` is that file.

Dotbot already has a canonical place for questions: the task itself.

- pending questions: `extensions.runner.pending_question` / `pending_questions`
- answered questions: `extensions.runner.questions_resolved`

`interview-answers.json` is specific to `start-from-prompt`. It exists so that workflow can reread answers later. That makes it workflow-owned state, not runtime-owned state.

So I don’t think we should add runtime exclusions, locks, merge handling, or accumulator behavior for this file. That only fixes this one workflow path and does not solve the broader shared-state problem.

I think the fix should be to change `start-from-prompt` to use the existing task-level question lifecycle. If it needs more answer detail, we should enrich `questions_resolved`, not duplicate answers into a shared product file.

Concretely:

1. stop writing `interview-answers.json` from runtime code
2. keep question/answer state on the task that asked the question
3. update `start-from-prompt` to read resolved answers from task state
4. avoid runtime locking/merge rules for this workflow-specific file

The conflict is real, but this PR feels like a band-aid over the wrong ownership boundary.

kabaogluemre:
@carlospedreira , thanks, you're right, and I've reworked the PR to fix the ownership boundary instead of patching the symptom.

carlospedreira (approved):

PR #529, opened 2026-06-22
feat(workflow): selective workflow re-run for completed runs

EnmaJim:
feat(workflow): selective workflow re-run for completed runs

## Linked issue

Closes #512

## Summary of changes

Adds a **targeted re-run** to the dashboard: from a finished workflow run, an
operator can re-run only selected tasks (and, by default, their downstream
dependents) **in place** — prior task outputs not in scope are preserved, and the
existing full-regeneration ("fresh") run is untouched. This reuses the runner's
existing behaviour: scoped to a run, it only picks up `todo` tasks, so resetting a
subset back to `todo` and relaunching the runner re-runs exactly that subset in
dependency order.

**Backend** (`src/runtime/Modules/Dotbot.Task/`)

- `Get-WorkflowTransitiveDependents` — pure reverse-dependency BFS over a run's
  task files; returns the transitive downstream dependents of the selected tasks
  (excludes `cancelled`, which severs the chain).
- `Reset-WorkflowTasksToTodo` — flips the targeted tasks back to `todo` via atomic
  writes, clearing `completed_at` and runner residue
  (`extensions.runner.pending_question` / `skip_reason` / `skip_detail`) while
  keeping the closed TaskInstance schema valid. Mirrors the trusted
  `Reset-InProgressTasks` crash-recovery path.
- Both functions exported from `Dotbot.Task.psd1` / `.psm1`.

**Endpoint** (`src/ui/server.ps1`)

- `POST /api/run/rerun` `{ run_id, task_ids[], target_only }`: validates the run
  and task ids, guards concurrency (refuses a still-active run; honours
  same-workflow policy) and the git/worktree precondition, computes the reset set
  (`task_ids` ∪ transitive dependents unless `target_only`), resets them, re-arms
  the run's live status to `running`, and relaunches the task-runner scoped to the
  same run. The fresh-run path (`/api/workflows/{name}/run`) is unchanged.

**Dashboard UI** (`src/ui/static/`)

- `modules/runs.js` — per-task checkboxes and a "Re-run selected" bar with an
  "include dependents" toggle, shown only on terminal runs; selection is preserved
  across state polls.
- `modules/controls.js` — `rerunSelectedTasks()` posts to the endpoint and
  refreshes state.
- `css/views.css` — styling for the new controls using existing design tokens.

## Screenshots / recordings

**Completed run — per-task checkboxes + "Re-run selected" + "include dependents"**

01-completed-run

**Task selected — button enabled**

02-tasks-selected

**After re-run (cascade) — run flips to RUNNING; selected task + dependents back to `todo`; unrelated task stays `done`**

03-rerun-cascade

**Dashboard context (Workflows tab)**

00-dashboard

## Testing notes

- All tests pass green.
- End-to-end testing was performed to confirm the feature works.

## Checklist

- [x] Tests added or updated
- [ ] Docs updated (if behaviour changed)
- [x] Linked issue exists

### DIFF
--- a/src/runtime/Modules/Dotbot.Task/Dotbot.Task.psd1
+++ b/src/runtime/Modules/Dotbot.Task/Dotbot.Task.psd1
@@ -26,6 +26,8 @@
         'Resolve-TaskReviewDecision'
         'Test-TaskCompletion'
         'Reset-InProgressTasks'
+        'Reset-WorkflowTasksToTodo'
+        'Get-WorkflowTransitiveDependents'
         'Reset-SkippedTasks'
         'Invoke-PostScript'
         'Invoke-PostScriptFailureEscalation'
--- a/src/runtime/Modules/Dotbot.Task/Dotbot.Task.psm1
+++ b/src/runtime/Modules/Dotbot.Task/Dotbot.Task.psm1
@@ -591,6 +591,189 @@ function Reset-InProgressTasks {
     return ,$resetTasks
 }
 
+function Get-WorkflowTransitiveDependents {
+    <#
+    .SYNOPSIS
+    Return the transitive downstream dependents of one or more tasks within a run.
+
+    .DESCRIPTION
+    Reads every task file in $RunDir, builds the reverse dependency graph from
+    each task's resolved 'dependencies' field, and walks it breadth-first from
+    $TaskIds. Returns the de-duplicated set of dependent task IDs (excluding the
+    seed IDs themselves). Tasks in 'cancelled' status are deliberately abandoned,
+    so they are never added to the set and the walk does not continue through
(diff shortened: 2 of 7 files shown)


carlospedreira:
@EnmaJim this is new feature territory and needs its own design/prioritization pass, so I’ll skip it for now rather than trying to solve it inside this PR.

At a design level, I’d prefer this to lean on the existing workflow-runner lifecycle instead of duplicating it. The re-run behavior should reopen selected tasks back to `todo`, then rely on the same runner/start/resume logic we already use elsewhere to move the run through `running`, failure handling, and completion. That keeps this from becoming a second, divergent workflow-start path that we have to maintain separately.

PR #530, opened 2026-06-23
fix(worktree): unblock squash-merge retry over EOL-stale untracked leftovers (#517)

kabaogluemre:
## Summary

Fixes #517 — a squash-merge apply that fails once and is retried would get
permanently parked in `needs-input` when the only obstacle was a leftover
untracked file from the prior attempt.

## Root cause

`Apply-TaskBranchPatch` guards untracked additions by comparing the local file
against the branch blob:

```powershell
$branchBlob = git rev-parse "$BranchName:$addedPath"      # LF-normalized blob
$localBlob  = git hash-object --no-filters -- $addedPath  # raw working-tree bytes
```

On Windows with `core.autocrlf=true`, the leftover written to the working tree by
a prior failed apply has **CRLF** line endings, while the branch blob is stored
**LF**. The `--no-filters` flag skips the clean filter, so the raw hash **never
matches** the LF blob — even when the content is byte-identical modulo EOL. The
guard then returns `Untracked file would be overwritten by task branch: <path>`
and the retry is blocked forever.

> This corrects the issue's original diagnosis ("the leftover always differs
> because it's stale"): the difference is line-ending normalization, not genuine
> content divergence.

## Why not just remove the guard

The issue proposed removing the blob-hash guard entirely and deleting the file
unconditionally. That would:
- break the deliberate regression test pinning the guard, and
- regress its purpose — silently overwriting a **genuinely divergent** untracked
  local file (e.g. a product artifact at a non-excluded path) with possible data
  loss.

## Fix

- **Drop `--no-filters`** so the comparison uses git's own normalization
  (clean filter + autocrlf). An EOL-only-stale leftover now hashes **equal** to
  the branch blob and is cleaned by the existing `Remove-Item` path → retry
  proceeds. Genuinely divergent content still differs → guard still blocks
  (protection intact, no silent overwrite).
- **Roll back apply artifacts on failure.** Because the guard returns early on any
  pre-existing divergent untracked file, once `git apply` runs every
  still-untracked file at an added path is *this attempt's* artifact. The failure
  branch now removes them so the retry starts from a clean slate (the caller's
  `git reset --hard HEAD` only handles tracked entries).

### DIFF
--- a/src/runtime/Modules/Dotbot.Worktree/Dotbot.Worktree.psm1
+++ b/src/runtime/Modules/Dotbot.Worktree/Dotbot.Worktree.psm1
@@ -1278,8 +1278,18 @@ function Apply-TaskBranchPatch {
             git -C $ProjectRoot ls-files --error-unmatch -- $addedPath 2>$null | Out-Null
             if ($LASTEXITCODE -eq 0) { continue }
 
+            # Compare against the branch blob using git's own normalization
+            # (clean filter + autocrlf), NOT raw bytes. On Windows with
+            # core.autocrlf=true a leftover from a prior failed apply is written
+            # to the working tree with CRLF while the branch blob is stored LF,
+            # so a raw (--no-filters) hash never matches even when the content
+            # is byte-identical modulo EOL — that false "divergence" permanently
+            # blocked squash-merge retries (issue #517). hash-object WITHOUT
+            # --no-filters yields the OID git would store, so an EOL-only-stale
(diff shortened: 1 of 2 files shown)


carlospedreira (approved):
Looks good. The fix keeps the divergent-untracked-file guard intact while making the stale-leftover comparison use Git's normal EOL normalization, and the regression coverage exercises both the Windows EOL retry case and the real-divergence block.

PR #539, opened 2026-06-25
feat(runtime): add task_append_evidence for on-demand evidence injection (#503)

EnmaJim:
## Linked issue

Closes #503

## Summary of changes

Adds a way to inject evidence into a specific **in-flight or queued task without
forcing it into `needs-input`**, so the task can re-poll its context and continue.
Until now the only attachment path was bound to the question/answer needs-input
flow.

- **New MCP tool `task_append_evidence`** (`src/mcp/tools/task-append-evidence/`):
  attaches an inline `note` and/or file `attachments` (base64) to any task,
  identified by `task_id`, with a `label` and optional `evidence_type`. Mirrors
  `task-create`; forwards to the runtime with the MCP actor.
- **New runtime endpoint `POST /tasks/<id>/evidence`**
  (`Invoke-AppendTaskEvidenceHandler` in `HttpServer.psm1`): under the task mutex,
  it decodes/writes attachments to
  `.bot/workspace/attachments/<taskId>/evidence/<evidenceId>/` (same allowed
  extensions as the answer flow, but decoupled from any question), **appends** an
  entry to `extensions.evidence`, writes the task atomically, and emits a
  `task_updated` activity event. It **never changes the task status** — so the
  task is not parked.
  - A dedicated endpoint (not PATCH) because PATCH's deep-merge _replaces_ arrays
    (lossy for appends) and the base64 decode/file-write belongs server-side.
- **Re-poll surfaces evidence**: `GET /tasks/<id>/context` (the existing
  `task_get_context` tool) now includes an `evidence` field, so a running agent
  picks up new evidence on its next poll without stopping.
- **Answer/inbox flow unchanged** (AC#3): evidence lives in its own open
  `extensions.evidence` namespace; `extensions.runner`, NotificationPoller,
  InboxWatcher, `Submit-TaskAnswer`, and `task-update` are untouched. No
  `TaskInstance` schema change (extensions namespace is open).

## Testing notes

- All tests green (`pwsh tests/Run-Tests.ps1`).
- New tests added: `task_append_evidence` tool contract (Test-McpSurface) and a
  live-runtime integration suite (Test-Runtime) covering append, status-unchanged,
  context re-poll, append-not-replace, runner-untouched, and validation/404.

## Checklist

- [x] Tests added or updated
- [x] Docs updated (if behaviour changed) — tool/endpoint self-documented; no user docs needed
- [x] Linked issue exists

### DIFF
--- a/src/mcp/tools/task-append-evidence/metadata.json
+++ b/src/mcp/tools/task-append-evidence/metadata.json
@@ -1,0 +1,40 @@
+{
+  "name": "task_append_evidence",
+  "description": "Attach structured evidence (an inline note and/or file artifacts) to any queued or in-flight task WITHOUT changing its status, so the task can re-poll its context via task_get_context and incorporate the new context without pausing. Decoupled from the needs-input question/answer flow; evidence is recorded under extensions.evidence and never touches extensions.runner. POST /tasks/<id>/evidence.",
+  "inputSchema": {
+    "type": "object",
+    "additionalProperties": true,
+    "properties": {
+      "task_id": {
+        "type": "string",
+        "description": "Canonical task ID (t_XXXXXXXX) to attach evidence to."
+      },
+      "label": {
+        "type": "string",
+        "description": "Short human-readable label for this evidence batch."
+      },
+      "evidence_type": {
+        "type": "string",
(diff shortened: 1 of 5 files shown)


carlospedreira:
@EnmaJim this is new feature territory and needs its own design/prioritization pass, so I’ll skip it for now rather than trying to land a new in-flight task input path directly from this PR.

PR #538, opened 2026-06-25
feat(runtime): gate task execution on input/output artifacts (#510)

EnmaJim:
## Linked issue

Closes #510

## Summary of changes

Adds a **task input/output artifact contract** so a task can no longer be marked
`done`, nor silently feed a broken context downstream, when a required artifact
is absent, empty, or malformed.

- **Shared predicate `Test-DotbotArtifact`** (`src/runtime/Scripts/Invoke-WorkflowProcess.ps1`):
  `exists → non-empty → (for .json) parses`. Minimal, toolchain-free; Markdown/
  source bottom out at the non-empty check by design.
- **Producer-exit gate strengthened**: `Test-TaskOutput` now rejects empty/
  malformed declared outputs (previously existence-only), satisfying the issue's
  "absent, empty, or fails shape validation".
- **Consumer-entry gate added**: new `Test-TaskInput` validates a task's declared
  `inputs` before it runs. On failure the task is escalated to `needs-input` with
  a specific, actionable message (e.g. `Required input 'briefing/repo-scan.md' is absent`)
  via the existing `Set-WorkflowTaskNeedsInput`. The gate runs **before** any
  provider session, so a blocked task wastes no tokens.
- **New first-class `inputs` task field** threaded through the closed schemas and
  builders (`TaskDefinition`, `TaskInstance`/`New-TaskInstance`, `New-WorkflowTask`,
  `_FlattenTask`) plus API/MCP parity (`HttpServer` POST/PATCH `/tasks`,
  `task-create` MCP tool).
- **Downstream blocking + resume** reuse existing machinery: a `needs-input` task
  isn't reselected and its `depends_on` dependents stay blocked; answering
  "retry" requeues it and the gate re-validates — no full workflow restart.
- Tasks without `inputs`/`outputs` are unaffected (existing behaviour preserved).

## Screenshots / recordings

**Required input artifacts surfaced in ACTION REQUIRED, with retry/skip:**

Action required slideout showing missing input artifacts with retry/skip options

**Resume: after answering "Investigate logs and retry", the task requeues with no error and runs to done:**

Task resuming after the retry answer with no error

## Testing notes

- All tests green (`pwsh tests/Run-Tests.ps1`).
- New tests added for the input/output gate, the `inputs` schema, and gate wiring.

## Checklist

- [x] Tests added or updated
- [x] Docs updated (if behaviour changed) — schema doc-comments updated; no user docs needed
- [x] Linked issue exists
[...]

PR #540, opened 2026-06-25
fix(worktree): claim non-prompt tasks before worktree creation to prevent duplicate-worktree race

IBondarenko-iwg:
## Linked issue

Closes #536

## Summary of changes

Non-prompt auto-dispatch tasks (`script`, `mcp`, `task_gen`) claimed in-progress status **after** worktree creation. Two concurrent runners could both observe the same task as `todo` via the lock-free `Get-NextWorkflowTask`, both call `git worktree add` to the same path, and the second would fail with `fatal: '...' already exists`.

**Root fix:** Added a pre-worktree claim guard for non-prompt tasks mirroring the existing prompt-task guard. Calls `Invoke-TaskMarkInProgress` (HTTP POST with server-side mutex) before any worktree I/O. `no_op=true` in the response signals the task was already claimed — the guard fetches a replacement task and breaks to the outer loop so full `task_gen`/prompt-template recovery runs on it. Removed the post-worktree `Set-TaskInProgressForExecutorDispatch` call, now superseded.

**Additional hardening (from code review):**
- Re-syncs `$processData.task_id` / `$env:DOTBOT_CURRENT_TASK_ID` after the guard in case `$task` was replaced during retry
- Fatal errors (unexpected status, non-HTTP exceptions) re-throw instead of silently retrying
- 2s back-off before outer-loop `continue` when all claim attempts exhaust
- On task swap, always breaks to outer loop so the replacement goes through `task_gen`/`prompt_template` recovery
- Updated `Test-ProcessDispatch.ps1` assertion to match `Invoke-TaskMarkInProgress` instead of the removed function name

## Testing notes

1. **Unit** — `pwsh tests/Test-ProcessDispatch.ps1`: the "legal executor status transitions" assertion now validates the claim guard is present.
2. **Race reproduction** — start two task-runner processes against a workflow with at least one pending non-prompt task. Before this fix both runners race to `git worktree add`; after, the second detects `no_op=true` and picks up a different task.
3. **Standalone runner** — verify a single runner (`-Slot` not set) still claims and executes non-prompt tasks normally; the guard is unconditional and must not regress the single-runner path.

## Checklist

- [x] Tests added or updated
- [x] Linked issue exists

### DIFF
--- a/src/runtime/Scripts/Invoke-WorkflowProcess.ps1
+++ b/src/runtime/Scripts/Invoke-WorkflowProcess.ps1
@@ -1375,7 +1375,8 @@ try {
         # --- Multi-slot claim guard ---
         # When running with -Slot (concurrent workflow processes), another slot may
         # have claimed this task between our Get-NextWorkflowTask and this point.
-        # Only needed for prompt tasks — non-prompt tasks are guarded by the slot 0 check above.
+        # Only needed for prompt tasks — non-prompt tasks have their own claim guard
+        # before worktree creation below.
         if ($Slot -ge 0 -and $taskTypeCheck -eq 'prompt') {
             $claimOk = $false
             for ($claimAttempt = 0; $claimAttempt -lt 5; $claimAttempt++) {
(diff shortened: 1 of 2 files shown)


IBondarenko-iwg:
@carlospedreira Heads up: potential merge conflict with **#538** (`feature/issue-510`). Both PRs modify `Invoke-WorkflowProcess.ps1` at the same hunk (~line 1485). This PR removes `Set-TaskInProgressForExecutorDispatch` and inserts the claim guard block there; #538 also adds lines in that area. No semantic overlap — whoever merges second will need a manual hunk combine.

carlospedreira (approved):

PR #573, opened 2026-07-02
fix(process): dep-gated conditions + barriers wait for spawned children (#569)

IBondarenko-iwg:
## Linked issue

Closes #569
Refs #557

> Child PR of the 8-bug epic #557, delivering **bugs 4 and 5**. Merging `Closes` #569; #557 stays open until the last child lands.

## Summary of changes

`src/runtime/Modules/Dotbot.Process/Dotbot.Process.psm1` — both fixes in the `Get-NextWorkflowTask` candidate loop:

**Bug 4 — premature, cascading condition-skip.** The loop evaluated `_IsManifestConditionMet` *before* `_AreDepsMet`, so a task whose manifest `condition` points at a file its own upstream dependency produces was marked `skipped`/`condition-not-met` at launch (the file doesn't exist yet), added to the done-set, and — since `condition-not-met` satisfies dependents — cascaded the skip down the chain (observed live: 6 implementation-chain tasks skipped in `start-from-jira`). Now `_AreDepsMet` is checked **first**: a task with unmet deps stays `todo` (blocked, never marked), and because `Get-NextWorkflowTask` reloads from disk each call, its condition re-evaluates once the producer completes. A genuinely-false condition on a dep-met task is still a real `condition-not-met` skip.

**Bug 5 — barriers ignored dynamically-spawned children.** `_AreDepsMet` only inspects the explicit `dependencies` array, so a `barrier` fired the instant its generator dependency went `done` — while the children the generator spawned into `tasks/todo` (stamped `extensions.runner.generated_by` / `provenance.expanded_by`) were still running. New `_HasOutstandingGeneratedChildren` blocks a `type=barrier` candidate while any non-complete task was generated by one of its dependencies. "Complete" reuses the **same rule as the dependency done-set** (done/cancelled/split, or `skipped` only with an intentional reason), so a framework-error skip (`max-retries`, which `Reset-SkippedTasks` retries) or a `needs-input`/`needs-review` child keeps the barrier waiting rather than firing over unfinished work. Barrier-scoped, direct children only — verified complete for all shipped workflows; widening is a one-line scope-predicate change.

## Testing notes

`tests/Test-TaskActions.ps1` (Layer 2) adds behavioral coverage calling the exported `Get-NextWorkflowTask` against on-disk task fixtures:
- **Bug 4:** producer + a dependent conditioned on the producer's output. Pass 1 (file absent): producer is selected and the conditioned task **stays `todo`** (not `condition-not-met`). Mark producer done + create the file → pass 2 selects the conditioned task.
- **Bug 5:** `done` generator + a `todo` child stamped `generated_by`, plus a higher-priority `barrier` depending on the generator. Assert the barrier is **not** selected while the child is non-terminal (child selected); mark child `done` → barrier is then selected.

Run: `pwsh ./tests/Test-TaskActions.ps1` (or `./tests/Run-Tests.ps1 -Layer 2`). `PARSE OK`; PSScriptAnalyzer clean on the changed region; reviewed with `/pwsh-review` (diff-bug + idioms + history) — one major (child-complete semantics) applied, no blockers.

## Checklist

- [x] Tests added or updated
- [x] Linked issue exists

### DIFF
--- a/src/runtime/Modules/Dotbot.Process/Dotbot.Process.psm1
+++ b/src/runtime/Modules/Dotbot.Process/Dotbot.Process.psm1
@@ -623,6 +623,51 @@ function Get-NextWorkflowTask {
         }
         return $true
     }
+    function _HasOutstandingGeneratedChildren { param($Task, $All)
+        # True when any non-terminal task was generated/expanded by one of $Task's
+        # dependencies. Children stamp extensions.runner.generated_by = <parent id>
+        # and provenance.expanded_by = "task:<parent id>"; a barrier's dependencies
+        # are canonical parent ids, so this is a direct id match (with a slug
+        # fallback for name-based deps). Scope note: this is invoked only for
+        # type=barrier tasks and matches direct children (not nested descendants);
+        # widening is a one-line change at the call site.
+        $deps = @($Task.dependencies) | Where-Object { $_ } | ForEach-Object { [string]$_ }
+        if ($deps.Count -eq 0) { return $false }
+        $depSet = [System.Collections.Generic.HashSet[string]]::new([System.StringComparer]::OrdinalIgnoreCase)
+        foreach ($d in $deps) {
+            [void]$depSet.Add($d)
+            $slug = ($d -replace '[^a-zA-Z0-9\s-]','' -replace '\s+','-').ToLowerInvariant()
+            if ($slug) { [void]$depSet.Add($slug) }
+        }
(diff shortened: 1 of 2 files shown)


carlospedreira (approved):
Reviewed locally and via connector. Focused regression suite passed: tests/Test-TaskActions.ps1 reported 186 passed, 0 failed. No blocking issues found.

PR #637, opened 2026-07-10
fix(process): pin process-state writes to canonical BotRoot, not worktree

emrahzunicaplab:
## Linked issue

Closes #612

## Summary of changes

`Write-ProcessFile` and `Test-ProcessStopSignal` resolve `BotRoot` via `$PWD`
whenever no `-BotRoot` is passed. During task execution,
`Invoke-WorkflowProcess.ps1` `Push-Location`s into the task's worktree, whose
`.bot/.control` is a junction back to the canonical `.control` dir. Worktree
teardown (`Complete-TaskWorktree`, `Reset-TaskWorktree`, `Remove-OrphanWorktrees`)
removes that junction before removing the worktree itself, so a heartbeat write
racing the teardown resolves to a path that no longer exists and silently drops
after exhausting retries (`Could not find a part of the path '…\worktrees\1\task-…\.bot\.control\processes\proc-….json.tmp'`).

Changes:

- **Pin every process-registry write inside the worktree-scoped execution window**
  to `$botRoot` — the canonical root captured once at process start and never
  reassigned. Covers all `Write-ProcessFile` / `Test-ProcessStopSignal` calls
  between the worktree `Push-Location` and its `Pop-Location`.
- **Thread the same canonical root through `Invoke-TaskClarificationLoopIfPresent`**
  via a new `-ProcessBotRoot` parameter, kept deliberately separate from its
  existing worktree-scoped `-BotRoot` (which is still used for the answers-file
  path).
- **Add a `processes/` directory guard to `Write-ProcessFile`**, mirroring
  `Write-ProcessActivity`, so a missing directory self-heals instead of failing
  outright.

Process writes outside the worktree window (CWD = project root) are intentionally
left unpinned — they already resolve to the canonical root. Sibling runners
(`Invoke-PromptProcess.ps1`, `Invoke-DotbotProcess.ps1`) never `Push-Location`
into a worktree, so they are unaffected. Activity-log writes go through
`Write-BotLog`, which is init-pinned to the canonical control dir at startup, so
they were never vulnerable to this race.

## Testing notes

- `pwsh` parse check on both changed files — clean.
- `tests/Test-ProcessRegistry.ps1` — 20/20 pass.
- `tests/Test-Compilation.ps1` — 372 pass, 0 fail.
- Reproduced the exact #612 scenario in a sandbox with a real `.control` junction:
  a `BotRoot`-pinned write reaches the canonical `processes/` dir and survives
  junction teardown, the directory guard recreates a deleted `processes/`, while
  an unpinned write (the old behaviour) resolves to the now-orphaned worktree path
  and never reaches canonical — confirming both the bug and the fix.

## Checklist
- [x] Linked issue exists

### DIFF
--- a/src/runtime/Modules/Dotbot.Process/Dotbot.Process.psm1
+++ b/src/runtime/Modules/Dotbot.Process/Dotbot.Process.psm1
@@ -106,6 +106,9 @@ function Write-ProcessFile {
         [string]$BotRoot
     )
     $processesDir = Get-ProcessesDir -BotRoot $BotRoot
+    if (-not (Test-Path $processesDir)) {
+        New-Item -Path $processesDir -ItemType Directory -Force | Out-Null
+    }
     $filePath = Join-Path $processesDir "$Id.json"
     $tempFile = "$filePath.tmp"
     $retry = Get-ProcessRetryConfig -BotRoot $BotRoot
--- a/src/runtime/Scripts/Invoke-WorkflowProcess.ps1
+++ b/src/runtime/Scripts/Invoke-WorkflowProcess.ps1
@@ -837,7 +837,12 @@ function Invoke-TaskClarificationLoopIfPresent {
         [string]$ModelName,
         [bool]$ShowDebug,
         [bool]$ShowVerbose,
-        [string]$PermissionMode
+        [string]$PermissionMode,
+        # Canonical project .bot root for process-registry writes (proc-*.json).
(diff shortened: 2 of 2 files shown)


carlospedreira (approved):
Looks good. I reviewed the process-registry/worktree-rooting changes and ran focused validation against the PR commit: parser check on the changed PowerShell files plus `tests/Test-ProcessRegistry.ps1` (20/20 passing). I don’t see a merge blocker.

PR #660, opened 2026-07-20
fix(process): reap orphaned process tree via kill-on-close Job Object (#645)

emrahzunicaplab:
## Linked issue

Closes #645

## Summary of changes

On an **abrupt** worker death (console close, `Stop-Process -Force`, crash), the
in-process `finally` cleanup in the harness adapters never runs, so the worker's
child subtree (`claude.exe` → MCP server → tool children, plus git/preflight
helpers) is orphaned and keeps running — consuming tokens.

This binds each worker's whole subtree to the worker's own lifetime via a Windows
**kill-on-close Job Object**:

- New `Register-DotbotKillOnCloseJob` in `Dotbot.Process` creates a job with
  `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE`, assigns the current process to it, and
  holds the sole job handle open for the process lifetime. Child processes inherit
  job membership, so whenever the worker dies the OS closes the handle and the
  kernel terminates the whole subtree.
- `Invoke-DotbotProcess.ps1` calls it right after importing `Dotbot.Process`.
- Strictly **best-effort** and a **backstop**, not a replacement: graceful teardown
  still uses the existing `finally` cleanup. On non-Windows (children re-parent to
  init and are reaped via `pkill`) or on any failure, it logs at Debug and returns
  `$false`; callers must never fail because of it. Idempotent across repeated
  calls and `Import-Module -Force`.
- Failure paths log the captured `Win32` error code to aid field diagnosis.

## Testing notes

- `pwsh tests/Test-ProcessRegistry.ps1` — adds an end-to-end case that binds a
  worker, spawns a `mid → leaf` subtree (modelling `claude.exe → MCP server`),
  force-kills the worker **without** running any cleanup, and asserts the kernel
  reaps both descendants. Non-Windows asserts the documented `$false` no-op.
- Full Layer 1–3 suite is green apart from the pre-existing, unrelated
  `New-TaskWorktree` Windows long-path failure (`Filename too long`), which also
  fails on `main` and is untouched by this PR.

## Checklist

- [x] Tests added or updated
- [x] Docs updated (if behaviour changed) — n/a (internal backstop; documented in-code)
- [x] Linked issue exists

### DIFF
--- a/src/runtime/Modules/Dotbot.Process/Dotbot.Process.psd1
+++ b/src/runtime/Modules/Dotbot.Process/Dotbot.Process.psd1
@@ -27,6 +27,8 @@
         'Test-WorkflowComplete'
         # Child process spawning
         'Start-DotbotChildProcess'
+        # Process-tree lifetime binding
+        'Register-DotbotKillOnCloseJob'
     )
 
     CmdletsToExport   = @()
--- a/src/runtime/Modules/Dotbot.Process/Dotbot.Process.psm1
+++ b/src/runtime/Modules/Dotbot.Process/Dotbot.Process.psm1
@@ -944,6 +944,170 @@ function Start-DotbotChildProcess {
 
 #endregion
 
+#region Kill-on-close Job Object (Windows)
+
+# Sole handle to the process's kill-on-close job, held open for the whole
+# process lifetime. IntPtr::Zero means "not bound yet". Never CloseHandle this:
+# the OS closing it on process death is precisely what reaps the subtree.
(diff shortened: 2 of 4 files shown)


carlospedreira (approved):
Reviewed the Windows Job Object lifecycle, handle ownership, cleanup behavior, cross-platform guard, surrounding process code, and regression coverage. The unchanged head is mergeable; required CI passes across Windows, macOS, and Ubuntu. Follow-up-level test cleanup/documentation concerns are accepted and do not block this fix.

PR #661, opened 2026-07-21
fix: resolve dependencies via registry Machine/User PATH on Windows(#504)

EnmaJim:
## Linked issue

Closes #658

## Summary of changes

On Windows, dotbot reported `git: MISSING` / `claude: MISSING` and the workflow child window flashed and died whenever the process inherited a PATH missing the Machine or User registry scope (e.g. Git installed system-wide, Claude installed per-user) — even though both tools worked in the user's terminal. Every detection site used `Get-Command`, which only searches the process PATH.

**Fix:**

- New in `Dotbot.Core` (dependency-free, imported first by every entry point):
  - `Resolve-DotbotExternalCommand` — resolves a command via the process PATH, falling back to a registry `Machine` → `User` PATH scan on Windows; reports which scope the tool was found in and can repair the session PATH.
  - `Repair-DotbotProcessPath` — one-time, append-only merge of registry PATH directories missing from `$env:PATH`; opt-out via `DOTBOT_SKIP_PATH_REPAIR`; no-op on non-Windows.
- `Test-Preflight` and the `Test-GitReadyForWorktree` git gate resolve through the new resolver with session repair. Registry-scope hits produce an actionable check line naming the tool, the PATH scope, the directory, and the permanent fix. The exact `MISSING` strings are unchanged for truly-absent tools.
- The three entry-point preambles (`bin/dotbot.ps1`, `Invoke-DotbotProcess.ps1`, `src/ui/server.ps1`) call `Repair-DotbotProcessPath` once, transparently healing all downstream spawn sites (harness adapters, `ProcessStream`) and UI capability probes without modifying them.
- `dotbot doctor` detects the split instead of healing it (the CLI preamble deliberately skips repair for `doctor` so it observes the raw inherited PATH): a registry-only hit emits a `split PATH detected` **Warn** plus session and permanent fix commands.
- Secondary bug fixed: `bin/dotbot.ps1` swallowed `doctor.ps1`'s exit code — `dotbot doctor` exited 0 even with failing checks. The dispatcher now propagates it (same pattern as `dotbot run`).

## Testing notes

`tests/Test-DependencyPathResolution.ps1` encodes the issue's acceptance criteria and was failing 11/18 before the fix — it now passes **18/18**. It simulates the split by spawning child pwsh processes with tool directories stripped from the process PATH while the registry PATH stays intact, and covers: preflight resolution via Machine and User scopes, the actionable scope-naming messages, doctor's split detection, and doctor exit-code propagation (with two scenarios that stay non-zero even after the PATH fix, so the propagation assertion can't trivially pass).

- `pwsh -NoProfile -ExecutionPolicy Bypass -File tests/Test-DependencyPathResolution.ps1` → 18 passed, 0 failed.
- `pwsh tests/Run-Tests.ps1 -Layer all` → Layers 1–3 all passed (including the static scanners and both Layer-3 mock-provider suites; the append-only repair preserves mock-first PATH ordering).
- Live verification on a real split-PATH machine (git in Machine PATH, claude in User PATH): preflight passes from stripped shells naming the correct scope; post-preflight `Get-Command` succeeds in the same process (session-repair proof); a directory freshly added to the real User PATH is picked up; removing the tool everywhere still yields the exact `MISSING` failure; end-to-end `dotbot run smoke-test` from a claude-stripped shell survives preflight and runs, while `dotbot doctor` from the same shell reports the split and exits 1.

## Checklist

- [x] Tests added or updated
- [ ] Docs updated (if behaviour changed)
- [x] Linked issue exists

### DIFF
--- a/bin/dotbot.ps1
+++ b/bin/dotbot.ps1
@@ -85,6 +85,14 @@ foreach ($arg in $RawArgs) {
 $Command = $FilteredArgs[0]
 [array]$SubArgs = if ($FilteredArgs.Count -gt 1) { $FilteredArgs[1..($FilteredArgs.Count-1)] } else { @() }
 
+# Merge registry Machine/User PATH into this process (Windows split-PATH fix).
+# Skipped for 'doctor': doctor must observe the raw inherited PATH to detect
+# and report the split. Guarded because a project-vendored runtime may ship an
+# older Dotbot.Core without the repair function.
+if ($Command -ne 'doctor' -and (Get-Command Repair-DotbotProcessPath -ErrorAction SilentlyContinue)) {
+    $null = Repair-DotbotProcessPath
+}
+
 # Convert CLI args to a hashtable for proper named-parameter splatting.
 # Array splatting only does positional binding; hashtable splatting is
 # required for named parameters like -Workflow / -Stack.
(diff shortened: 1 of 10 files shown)


carlospedreira (approved):
Reviewed the rebased final head. PATH repair opt-out, split-PATH doctor behavior, deterministic scope/elevation coverage, executable precedence, and cross-platform behavior are addressed; local Layers 1-3 and all required CI checks pass.

PR #692, opened 2026-08-20
fix: repair CLI dispatch, runtime lifecycle and task worktree merges

### LINKED ISSUE #679: Task merges target the recorded `base_branch`, silently writing to the trunk or blocking the run
### What happened

`Complete-TaskWorktree` takes its merge target from the `base_branch` recorded in
`.bot/.control/worktree-map.json` when the task worktree was created, and never reconciles it with the
branch the operator's main checkout is actually on (`Dotbot.Worktree.psm1:1666`):

```powershell
$baseBranch = $entry.base_branch ?? (Resolve-MainBranch -ProjectRoot $ProjectRoot -BotRoot $BotRoot)
```

When the recorded value differs from the checked-out branch, one of two things happens — and which one
depends on whether anything happens to block `git checkout`:

**Case A — nothing blocks it: the task's work is silently committed to the trunk.**

```
success       : True
failure_kind  :
message       : Squash-merged to master and cleaned up
```

```
master                       18255dc -> f5a8fd7   feat: RM-B0 probe task [task:t_RMB0te]
                                                  apps/web/src/rm-b0-marker.ts | 1 +
workflow/plan-feature-RMB0   unchanged at 18255dc, file absent
HEAD                         workflow/plan-feature-RMB0 -> master   (left there)
```

The commit landed on `master` while the operator was on `workflow/plan-feature-RMB0`. No warning, no
non-zero status. The branch the operator was on received nothing. The working copy was left on
`master`.

**Case B — something blocks the checkout: the run halts with an unexplained error.**

```
failure_kind : exception
message      : Error during merge: Failed to checkout master in <repo> (currently on: workflow/plan-feature-RMB1)
at Assert-OnBaseBranch,   ...\Dotbot.Worktree.psm1: line 245
at Complete-TaskWorktree, ...\Dotbot.Worktree.psm1: line 1754
```

The most reachable blocker is Dotbot's own pre-merge stash. `Dotbot.Worktree.psm1:1744-1748` excludes
`.bot/workspace/decisions/` from the stash:

```powershell
git -C $ProjectRoot stash push -u -m "dotbot-pre-merge-$TaskId" -- `
    '.' ':!.bot/workspace/tasks/' ':!.bot/workspace/decisions/'
```

…but that tree **is tracked** (`init-project.ps1:230-245` seeds `.gitkeep` files into
`decisions/{accepted,deprecated,proposed,superseded}`; `.bot/.gitignore` does not cover it), it is
**committed** by this very function at `:1880`, and unlike `.bot/workspace/tasks/` — scrubbed at
`:1738-1739` — it is **never scrubbed**. So a dirty decisions file survives the stash and blocks the
checkout the stash exists to enable. `:1750-1752` states that purpose explicitly.

Once it fails, recovery is impossible without manual surgery: `New-TaskWorktree:1429-1432` writes a map
entry only `if (-not $lockedMap.ContainsKey($TaskId))`, and
`Invoke-WorkflowProcess.ps1:130-139` early-returns when the worktree path still exists. `base_branch`
is therefore **never refreshed**, so answering the escalation's "A — investigate and retry" replays the
identical failure indefinitely.

### What you expected

A task's work should be merged into the branch the run is actually building on, and the operator's
working copy should not be moved off their branch or written to without intent.

Concretely:

1. If the recorded `base_branch` no longer matches the run's actual base, Dotbot should reconcile it or
   refuse with a message naming both branches — not silently commit to the trunk.
2. The pre-merge stash must include every **tracked** path it needs to move away from, or scrub
   `.bot/workspace/decisions/` the way it already scrubs `.bot/workspace/tasks/`.
3. A retry after a merge failure should re-resolve `base_branch` rather than reuse the stale value.

### Steps to reproduce

Deterministic, no AI provider needed. Tested on Windows 11 / pwsh 7.6.4 / git 2.54.0.

```powershell
$env:DOTBOT_HOME = '<dotbot checkout>'
$repo = 'C:\tmp\repro'; $bot = "$repo\.bot"

# 1. a repo with a trunk, dotbot initialised and committed
git init -b master $repo; cd $repo
git config user.email a@b.c; git config user.name t
'x' > README.md; git add -A; git commit -qm init
& $env:DOTBOT_HOME\bin\dotbot.ps1 init -y
git add -A; git commit -qm 'chore: dotbot init'

Import-Module $env:DOTBOT_HOME\src\runtime\Modules\Dotbot.Core\Dotbot.Core.psm1 -Force -DisableNameChecking
Import-Module $env:DOTBOT_HOME\src\runtime\Modules\Dotbot.Settings\Dotbot.Settings.psd1 -Force -DisableNameChecking
Import-Module $env:DOTBOT_HOME\src\runtime\Modules\Dotbot.Worktree\Dotbot.Worktree.psd1 -Force -DisableNameChecking
New-Item -ItemType Directory -Force "$bot\.control" | Out-Null

# 2. a task worktree whose recorded base_branch is master, with a real change committed
$wt = New-TaskWorktree -TaskId t_probe -TaskName 'probe' -ProjectRoot $repo -BotRoot $bot -BaseBranch master
'must not land on master' > "$($wt.worktree_path)\marker.txt"
git -C $wt.worktree_path add -A; git -C $wt.worktree_path commit -qm 'feat: marker'

# 3. the operator moves the main checkout to another branch (as a workflow run does)
git -C $repo branch workflow/plan-feature-TEST master
git -C $repo checkout -q workflow/plan-feature-TEST

# ---- CASE A: clean tree -> silent merge into master ----
Complete-TaskWorktree -TaskId t_probe -ProjectRoot $repo -BotRoot $bot -SkipRemotePush
#   -> success=True, "Squash-merged to master and cleaned up"
git -C $repo log --oneline -1 master        # the task commit is on master
git -C $repo rev-parse --abbrev-ref HEAD    # master — moved off the operator's branch

# ---- CASE B: add the blocker -> the reported exception ----
# repeat steps 2-3 with a fresh task id, then before completing:
$dec = "$repo\.bot\workspace\decisions\proposed\dec-001.md"
'original' > $dec; git -C $repo add -f $dec; git -C $repo commit -qm 'docs: dec-001'
git -C $repo checkout -q workflow/plan-feature-TEST
'differs here' > $dec; git -C $repo add -f $dec; git -C $repo commit -qm 'docs: dec-001 differs'
'dirty and uncommitted' > $dec               # the blocker
Complete-TaskWorktree -TaskId t_probe2 -ProjectRoot $repo -BotRoot $bot -SkipRemotePush
#   -> failure_kind=exception, "Failed to checkout master in <repo> (currently on: workflow/plan-feature-TEST)"
```

The blocker mechanism in isolation, with no Dotbot at all:

```powershell
git stash push -u -m t -- '.' ':!.bot/workspace/tasks/' ':!.bot/workspace/decisions/'
#   No local changes to save        <- the only dirty file is the one it excludes
git checkout master
#   error: Your local changes to the following files would be overwritten by checkout:
#           .bot/workspace/decisions/proposed/dec-001.md
```

### Environment

```
OS: Windows 11 Pro 10.0.26200 | dotbot v4.0.2 (main @ 7c95b466)
pwsh 7.6.4 | git 2.54.0.windows.1 | node 24.16.0 | dotnet SDK 10.0.302
DOTBOT_HOME set explicitly; %APPDATA%\dotbot\user-settings.json absent
git config: init.defaultBranch=master, core.autocrlf=true
AI provider: not involved — reproduced through direct Dotbot.Worktree calls and raw git
Target repo: ~1129 files, tracked .bot/workspace, single remote
```

### Severity

high

### Logs / screenshots

Reported by a user on a real project, matching Case B exactly — same `failure_kind`, same message
template, same line numbers:

```
Merge failure kind: exception
Message: Error during merge: Failed to checkout master in
  C:\Users\<user>\source\repos\apms-performanceevaluation (currently on: workflow/plan-feature-TAjy)
Detail: Failed to checkout master in ... (currently on: workflow/plan-feature-TAjy)
  at Assert-OnBaseBranch,   ...\Dotbot.Worktree\Dotbot.Worktree.psm1: line 245
  at Complete-TaskWorktree, ...\Dotbot.Worktree\Dotbot.Worktree.psm1: line 1754
  at <ScriptBlock>, ...\src\runtime\Scripts\Invoke-WorkflowProcess.ps1: line 2289
  at <ScriptBlock>, ...\src\runtime\Scripts\Invoke-DotbotProcess.ps1: line 442
Worktree preserved at: C:\...\worktrees\apms-performanceevaluation\task-t_Hd7Qhc-scan-relevant-codebase
```

State left behind after Case B (a retry cannot succeed):

```
HEAD                  : workflow/plan-feature-RMB1     <- operator's branch not restored
worktree preserved    : True
task branch preserved : YES
dirty file            : M .bot/workspace/decisions/proposed/dec-001-example.md
worktree-map entry    : base_branch STILL "master"
leaked stashes        : 0
```

**Workaround that does work** (verified end to end): set the base branch explicitly in
`.bot/.control/settings.json`:

```json
{ "git": { "base_branch": "feature/my-work" } }
```

```
New-TaskWorktree      -> map base_branch = "feature/my-work"
Complete-TaskWorktree -> "Squash-merged to feature/my-work and cleaned up"
master unchanged; feature/my-work advanced
```

### LINKED ISSUE #680: `Assert-OnBaseBranch` discards git's error and cannot read `git.base_branch`
### What happened

Two defects in one 33-line function (`Dotbot.Worktree.psm1:217-249`). They are filed together because
they share a cause — the function's parameter list is missing what it needs — and a fix touches the
same lines.

**(1) git's explanation is thrown away (`:243`).**

```powershell
if ($currentBranch -ne $BranchName) {
    git -C $ProjectRoot checkout $BranchName 2>&1 | Out-Null      # <- stderr discarded
    if ($LASTEXITCODE -ne 0) {
        throw "Failed to checkout $BranchName in $ProjectRoot (currently on: $currentBranch)"
    }
}
```

Demonstrated with a dirty tracked file that blocks the checkout:

```
Assert-OnBaseBranch -ProjectRoot <repo> -BranchName master

THROWN: Failed to checkout master in <repo> (currently on: workflow/plan-feature-RMB1)
  mentions 'local changes'   : False
  mentions the blocking file : False
  mentions 'overwritten'     : False
```

git had already said precisely what was wrong:

```
error: Your local changes to the following files would be overwritten by checkout:
        .bot/workspace/decisions/proposed/dec-001-example.md
Please commit your changes or stash them before you switch branches.
Aborting
```

Because the thrown string is what `Complete-TaskWorktree` puts into `failure_detail`, and what the
runner then shows the operator in the `needs-input` escalation, **every possible blocker produces the
same unactionable message** — a dirty tracked file, a branch already checked out in a second linked
worktree (`fatal: '<b>' is already used by worktree at ...`), or a Windows file lock
(`error: unable to unlink ...: Permission denied`). The operator cannot tell which, so the recommended
"A — investigate and retry" has nothing to investigate.

**(2) The function cannot see `git.base_branch`, so failure paths abandon it (`:225-231`).**

```powershell
param(
    [Parameter(Mandatory)][string]$ProjectRoot,
    [string]$BranchName                       # <- no -BotRoot parameter exists
)
if (-not $BranchName) {
    $BranchName = Resolve-MainBranch -ProjectRoot $ProjectRoot     # <- no -BotRoot passed
}
```

`Resolve-DotbotBaseBranch` only consults `git.base_branch` when a `-BotRoot` is supplied
(`:162`), so this fallback can never honour the setting:

```
Resolve-DotbotBaseBranch -ProjectRoot <r> -BotRoot <r>\.bot  ->  feature/qa-work
Resolve-DotbotBaseBranch -ProjectRoot <r>                    ->  master      <- config invisible
Resolve-MainBranch       -ProjectRoot <r>                    ->  master
```

Three call sites in the workflow runner omit `-BranchName`, and all three sit in failure /
terminal-non-done / skipped cleanup paths — `Invoke-WorkflowProcess.ps1:1774`, `:2390`, `:2413`:

```powershell
try { Assert-OnBaseBranch -ProjectRoot $projectRoot | Out-Null } catch { Write-BotLog -Level Warn ... }
```

So on a configured project, the happy path merges into the configured branch but **any task failure
force-checks-out `main`/`master`**, silently moving the operator's working copy off the branch they
configured:

```
configured git.base_branch : feature/qa-work
HEAD before                : feature/qa-work
Assert-OnBaseBranch -ProjectRoot <repo>        # exactly what line 1774 calls
returned                   : master
HEAD after                 : master
```

`:2413` carries the comment `# Re-assert base branch after failed-task cleanup (Fix: wrong-branch
merge)` — a fix for wrong-branch merges that itself ignores the configured base branch.

### What you expected

1. When `git checkout` fails, the thrown message should carry git's stderr so the operator learns
   *why*. The escalation shown to a human is only as useful as this string.
2. `Assert-OnBaseBranch` should accept `-BotRoot` and pass it through to `Resolve-MainBranch`, so its
   no-`-BranchName` fallback honours `git.base_branch` like every other resolution site. The three
   runner call sites should pass the run's resolved base (or `-BotRoot`) rather than defaulting to
   `main`/`master`.

### Steps to reproduce

**(1) Discarded stderr**

```powershell
$env:DOTBOT_HOME = '<dotbot checkout>'
$repo = 'C:\tmp\repro2'
git init -b master $repo; cd $repo
git config user.email a@b.c; git config user.name t
'x' > README.md; git add -A; git commit -qm init
& $env:DOTBOT_HOME\bin\dotbot.ps1 init -y; git add -A; git commit -qm 'chore: dotbot init'

Import-Module $env:DOTBOT_HOME\src\runtime\Modules\Dotbot.Core\Dotbot.Core.psm1 -Force -DisableNameChecking
Import-Module $env:DOTBOT_HOME\src\runtime\Modules\Dotbot.Worktree\Dotbot.Worktree.psd1 -Force -DisableNameChecking

# a tracked file whose content differs between the two branches, dirty in the working tree
$f = "$repo\.bot\workspace\decisions\proposed\d.md"
'on master' > $f; git add -f $f; git commit -qm 'docs: d on master'
git checkout -q -b other master
'on other' > $f; git add -f $f; git commit -qm 'docs: d on other'
'dirty' > $f

try { Assert-OnBaseBranch -ProjectRoot $repo -BranchName master } catch { $_.Exception.Message }
#   -> "Failed to checkout master in <repo> (currently on: other)"   ... and nothing else
git checkout master     # <- the reason Dotbot swallowed
```

**(2) Config ignored on the fallback path**

```powershell
New-Item -ItemType Directory -Force "$repo\.bot\.control" | Out-Null
'{ "git": { "base_branch": "other" } }' | Set-Content "$repo\.bot\.control\settings.json"
Import-Module $env:DOTBOT_HOME\src\runtime\Modules\Dotbot.Settings\Dotbot.Settings.psd1 -Force -DisableNameChecking

git checkout -q other
Resolve-DotbotBaseBranch -ProjectRoot $repo -BotRoot "$repo\.bot"   # -> other   (honoured)
Assert-OnBaseBranch -ProjectRoot $repo                              # -> master  (ignored)
git rev-parse --abbrev-ref HEAD                                     # -> master  (yanked)
```

### Environment

```
OS: Windows 11 Pro 10.0.26200 | dotbot v4.0.2 (main @ 7c95b466)
pwsh 7.6.4 | git 2.54.0.windows.1
DOTBOT_HOME set explicitly; %APPDATA%\dotbot\user-settings.json absent
AI provider: not involved — reproduced through direct Dotbot.Worktree calls
```

### Severity

medium

### Logs / screenshots

The real-world consequence, from a user report — the entire diagnostic payload for a blocked run,
containing no root cause:

```
Merge failure kind: exception
Message: Error during merge: Failed to checkout master in
  C:\Users\<user>\source\repos\apms-performanceevaluation (currently on: workflow/plan-feature-TAjy)
Detail: Failed to checkout master in ... (currently on: workflow/plan-feature-TAjy)
  at Assert-OnBaseBranch,   ...\Dotbot.Worktree.psm1: line 245
  at Complete-TaskWorktree, ...\Dotbot.Worktree.psm1: line 1754
  at <ScriptBlock>, ...\Invoke-WorkflowProcess.ps1: line 2289
  at <ScriptBlock>, ...\Invoke-DotbotProcess.ps1: line 442
```

The reason, recoverable only by re-running the checkout by hand afterwards:

```
dirty working-tree paths: 1
  ~ .bot/workspace/decisions/proposed/dec-001-example.md
--- target 'master': 1 paths differ from HEAD; 1 of them are dirty ---
  WOULD BLOCK git checkout master:
    ! .bot/workspace/decisions/proposed/dec-001-example.md
```

A minimal fix for (1) is to capture the output rather than discard it:

```powershell
$checkoutOutput = git -C $ProjectRoot checkout $BranchName 2>&1 | Out-String
if ($LASTEXITCODE -ne 0) {
    throw "Failed to checkout $BranchName in $ProjectRoot (currently on: $currentBranch): $($checkoutOutput.Trim())"
}
```

### LINKED ISSUE #681: Worktree setup writes `.bot/workspace/tasks/` into the main repo's `.git/info/exclude`
### What happened

`Ensure-DotbotWorktreeExcludes` (`Dotbot.Worktree.psm1:538-579`) is called once, from `:800`, with a
**task worktree** path. It resolves its target file like this:

```powershell
$excludePath = git -C $WorktreePath rev-parse --git-path info/exclude
```

For a linked worktree, `--git-path info/exclude` resolves into the **shared common git directory** —
i.e. the main repository's `.git/info/exclude`, not anything worktree-local. It then writes a
marker-delimited block that includes the project's task tree:

```
# dotbot generated execution environment: start
.mcp.json
.claude/
.codex/
.opencode/
.agents/
.gemini/
.bot/.control
.bot/.control/
.bot/.handoffs
.bot/.handoffs/
.bot/workspace/tasks          <-- now ignored in the MAIN repo
.bot/workspace/tasks/
.bot/content/
.bot/hooks/
.bot/settings/
# dotbot generated execution environment: end
```

Confirmed on a test project — `.git/info/exclude` mtime matches the first worktree creation, and:

```
> git check-ignore -v .bot/workspace/tasks/workflow-runs/2026-08-05-smoke-test-a09g/run.json
.git/info/exclude:19:.bot/workspace/tasks/   .bot/workspace/tasks/workflow-runs/.../run.json
```

Three consequences follow.

**(1) It contradicts the documented contract.** `README.md:118` states verbatim:

> **Keep `.bot/` tracked in git.** The workspace tree (tasks/decisions/plans/product) is your team's
> audit trail; the project's `.gitignore` already covers the gitignored bits. Worktree state replays
> depend on `.bot/workspace/` being visible to git.

`README.md:105` and `:216` both annotate `workspace/` as `(tracked)`.

**(2) `git add` on that tree becomes a silent no-op**, which breaks Dotbot's own commit step.
`Complete-TaskWorktree:1880` stages task state with plain `git add` and no `-f`:

```powershell
git add .bot/workspace/tasks/ .bot/workspace/decisions/
```

Demonstrated:

```
> git add .bot/workspace/tasks/
The following paths are ignored by one of your .gitignore files:
.bot/workspace/tasks
hint: Use -f if you really want to add them.
> git diff --cached --name-only
(empty)                                   <- nothing staged

> git add -f .bot/workspace/tasks/standalone/probe-task.json
> git diff --cached --name-only
.bot/workspace/tasks/standalone/probe-task.json    <- only -f works
```

**(3) State accumulates invisibly and survives cleanup.** After four workflow runs,
`git status --porcelain` was **empty** while four run directories sat on disk. Only `--ignored`
revealed them, and `git clean -fd` cannot remove them (ignored paths need `-x`):

```
> git status --porcelain
(empty)
> git status --porcelain --ignored -- .bot/workspace/tasks
!! .bot/workspace/tasks/workflow-runs/2026-08-05-smoke-test-BMrY/
!! .bot/workspace/tasks/workflow-runs/2026-08-05-smoke-test-Ikka/
!! .bot/workspace/tasks/workflow-runs/2026-08-05-smoke-test-MhFg/
!! .bot/workspace/tasks/workflow-runs/2026-08-05-smoke-test-a09g/
```

Dotbot's own integrity module warns about exactly this failure mode —
`src\mcp\modules\FrameworkIntegrity.psm1:15-18`:

> If `.bot/` is gitignored (repo .gitignore, `core.excludesFile`, `info/exclude`, or a deep rule),
> every git-based check returns empty and **silently passes**.

Finally, `.git/info/exclude` is never committed, so this is invisible in code review, differs per
machine, and `dotbot doctor` does not mention it — doctor reported `PASS: 8  WARN: 0  FAIL: 0` in
exactly this state.

### What you expected

Creating a task worktree should not change ignore rules for the operator's main checkout, and
`.bot/workspace/tasks/` should stay tracked as documented.

If generated files must be hidden **inside** the worktree, write to a worktree-local exclude
(`.git/worktrees/<name>/info/exclude`) rather than the shared common dir, or scope the entries to the
paths that are genuinely generated (`.mcp.json`, `.claude/`, `.bot/.control`, `.bot/.handoffs`) and
leave `.bot/workspace/tasks/` alone. If the tree is genuinely meant to be untracked, the README and
the `git add` at `:1880` need to change to match.

### Steps to reproduce

```powershell
$env:DOTBOT_HOME = '<dotbot checkout>'
$repo = 'C:\tmp\exclude-repro'
git init -b master $repo; cd $repo
git config user.email a@b.c; git config user.name t
'x' > README.md; git add -A; git commit -qm init
& $env:DOTBOT_HOME\bin\dotbot.ps1 init -y
git add -A; git commit -qm 'chore: dotbot init'

# baseline: the task tree is tracked and not ignored
git check-ignore -v .bot/workspace/tasks     # -> no output
type .git\info\exclude                       # -> stock git comments only

# run any workflow so a task worktree gets created (smoke-test needs no AI provider)
& $env:DOTBOT_HOME\bin\dotbot.ps1 workflow add smoke-test -y
& $env:DOTBOT_HOME\bin\dotbot.ps1 run smoke-test --watch

# after: the main repo's exclude file has been rewritten
type .git\info\exclude                       # -> contains .bot/workspace/tasks/
git check-ignore -v .bot/workspace/tasks     # -> .git/info/exclude:NN:.bot/workspace/tasks/
git add .bot/workspace/tasks/                # -> "paths are ignored ... Use -f"
git status --porcelain                       # -> clean, even though run state exists on disk
git status --porcelain --ignored -- .bot/workspace/tasks   # -> !! the run directories
```

### Environment

```
OS: Windows 11 Pro 10.0.26200 | dotbot v4.0.2 (main @ 7c95b466)
pwsh 7.6.4 | git 2.54.0.windows.1
DOTBOT_HOME set explicitly; %APPDATA%\dotbot\user-settings.json absent
Target repo: 1129 files, .bot committed, single remote, master default
dotbot doctor in the affected state: PASS 8 / WARN 0 / FAIL 0
```

### Severity

high

### Logs / screenshots

`.git/info/exclude` mtime lines up with the third run's worktree creation, not with `dotbot init`:

```
.git\info\exclude  LastWriteTime  2026-08-05T11:53:23   (run 3 started 11:53:16)
dotbot init ran at 11:37
```

The single caller, which passes a worktree path — so the write to the *main* repo's file is incidental
rather than intended:

```powershell
# Dotbot.Worktree.psm1:799-801
$frameworkRoot = Get-DotbotWorktreeFrameworkRoot
Ensure-DotbotWorktreeExcludes -WorktreePath $WorktreePath
Remove-DotbotGeneratedPath -Path (Join-Path $WorktreePath '.gemini') -WorktreePath $WorktreePath
```

Manual remediation, per project, on every machine: delete the marker-delimited block from
`.git/info/exclude`, then `git add -f .bot/workspace/tasks/`.

### LINKED ISSUE #682: `dotbot run <workflow>` without `--watch` never starts the runtime and fails asynchronously
### What happened

`dotbot help` documents the verb plainly, with `--watch` presented as an optional monitoring
convenience:

```
run               Run/rerun a workflow
```

Running it without `--watch` reports success and exits 0:

```
> dotbot run smoke-test
› Minting WorkflowRun for 'smoke-test'...
    Run: wr_BMrYEikb → 2026-08-05-smoke-test-BMrY
› Creating 1 task(s) under the run...
    + Headless Smoke Check [t_fn0JIddl]
  ✓ Created 1 task(s) for smoke-test
› Launching workflow process...

  ✓ Workflow 'smoke-test' started.
    Use 'dotbot run smoke-test --watch' to run and monitor without opening the UI.

exit=0
```

The detached runner then immediately parks the task in `needs-input`:

```json
"status": "needs-input",
"pending_question": {
  "question": "Per-task failure for 'Headless Smoke Check'",
  "context": "Per-task exception: Dotbot runtime endpoint not available. Start the runtime with
              'dotbot go' or set DOTBOT_RUNTIME_URL + DOTBOT_RUNTIME_TOKEN."
}
```

Nothing runs. `.bot/.control/smoke-test.json` (the workflow's only output) is never written.

Compare the two transcripts of the same command on the same project. **With** `--watch`:

```
› Launching workflow process...
› Starting headless runtime...
  ✓ Runtime ready at http://127.0.0.1:53523/
  ✓ Workflow 'smoke-test' started.
  ...
  ✓ Workflow process completed.        exit=0, 22.7s, task done
```

**Without** `--watch`, the `› Starting headless runtime...` step never appears. The runtime is
auto-started only on the watch path, while the detached runner requires it unconditionally.

That a `--no-auto-runtime` flag exists at all implies auto-start is meant to be the default behaviour,
not a side effect of `--watch`.

This is easy to hit without meaning to: any invocation that fails to bind `--watch` falls into the same
hole. `dotbot run smoke-test --poll-interval-ms` (flag with a missing value) silently drops `--watch`
handling and launched a real run that failed this way — that is how this was first observed.

### What you expected

Either:

1. `dotbot run <workflow>` starts the runtime the same way `--watch` does (with `--no-auto-runtime` as
   the documented opt-out); or
2. if a runtime is genuinely required and cannot be auto-started on this path, the command refuses
   **synchronously** with a non-zero exit and the message that the task file ends up carrying —
   `Dotbot runtime endpoint not available. Start the runtime with 'dotbot go'` — instead of printing
   `✓ Workflow 'smoke-test' started.` and exiting 0.

Reporting success for a run that cannot make progress is the core problem; a script or CI step has no
way to detect it.

### Steps to reproduce

Deterministic, no AI provider needed — `smoke-test` is a single `script` task.

```powershell
$env:DOTBOT_HOME = '<dotbot checkout>'
$repo = 'C:\tmp\nowatch-repro'
git init -b master $repo; cd $repo
git config user.email a@b.c; git config user.name t
'x' > README.md; git add -A; git commit -qm init
& $env:DOTBOT_HOME\bin\dotbot.ps1 init -y
git add -A; git commit -qm 'chore: dotbot init'
& $env:DOTBOT_HOME\bin\dotbot.ps1 workflow add smoke-test -y

# 1. no --watch: prints success, exits 0
& $env:DOTBOT_HOME\bin\dotbot.ps1 run smoke-test
"exit=$LASTEXITCODE"

# 2. wait, then inspect the task
Start-Sleep -Seconds 20
Get-ChildItem .bot\workspace\tasks\workflow-runs -Recurse -Filter 't_*.json' |
  ForEach-Object { (Get-Content $_ -Raw | ConvertFrom-Json) |
    Select-Object id, status, @{n='ctx';e={$_.extensions.runner.pending_question.context}} }
#   -> status = needs-input, ctx = "Dotbot runtime endpoint not available..."
Test-Path .bot\.control\smoke-test.json      # -> False, the workflow produced nothing

# 3. contrast: the same command WITH --watch succeeds
& $env:DOTBOT_HOME\bin\dotbot.ps1 run smoke-test --watch
"exit=$LASTEXITCODE"                          # -> 0
Test-Path .bot\.control\smoke-test.json      # -> True
```

### Environment

```
OS: Windows 11 Pro 10.0.26200 | dotbot v4.0.2 (main @ 7c95b466)
pwsh 7.6.4 | git 2.54.0.windows.1 | claude CLI 2.1.222 (not exercised — script task only)
DOTBOT_HOME set explicitly; %APPDATA%\dotbot\user-settings.json absent
No dotbot go / serve running when the command was issued
```

### Severity

high

### Logs / screenshots

The failed run still creates and **pushes** an integration branch, so a run that accomplished nothing
leaves debris on the remote (see #683). From `.bot/.control/activity.jsonl`:

```
error  Mandatory task failed, stopping workflow: Headless Smoke Check
text   Integration branch workflow/smoke-test-YszU pushed to <origin>. Open a PR into master.
text   Process proc-194577 finished (stopped, tasks_completed: 0)
```

The generated handoff document, showing the task never reached a worktree:

```
## Identity
- Task: t_TWjZjyvu - Headless Smoke Check
- Run: wr_YszUcnx2
- Reason: workflow-escalation
- Worktree: C:\Users\EnmaLap\Desktop\tanda        <- the main checkout, not a task worktree
- Branch: master

## Context
Per-task exception: Dotbot runtime endpoint not available. Start the runtime with 'dotbot go'
or set DOTBOT_RUNTIME_URL + DOTBOT_RUNTIME_TOKEN.
```

Workaround: always pass `--watch`, or start `dotbot go` before `dotbot run`.

### LINKED ISSUE #683: A failed task discards commits the agent already made; every run leaves an unreapable branch
### What happened

Two related lifecycle problems, both about work and branches that outlive — or fail to outlive — a run.

**(1) A failed task's committed work is discarded and is not in the reflog.**

With `quality_gate.enabled = true` and a deliberately failing `test_command`, a real agent run created
its file, **committed it**, then called `task_set_status(done)`. The gate correctly refused:

```
mcp__dotbot__task_set_status: MCP error -32603: Tool execution failed: validation error:
Hook 'enter-done' aborted the transition: Verify '05-verify-quality.ps1' failed:
Quality gate failed: test (exit 1)
```

Task ended `failed`. Then the worktree and task branch were deleted, taking the commit with them:

```
qg-probe.ts present on:
  master                     : False
  develop                    : False
  feature/qa-work            : False
  workflow/fast-prompt-duJ1  : False     (integration branch — correct for a failure)
worktrees preserved : 0
task branch         : gone
```

Recoverability:

```
git reflog --all | grep qg-probe        -> not in reflog     (branch deleted, its reflog went too)
git stash list                          -> (empty)
git fsck --lost-found                   -> 8 dangling commits, one of which is the work:
  55f3b6b  feat: add QG_PROBE constant for quality gate probe
```

So the code exists only as a dangling git object, outside the reflog, and the next `git gc` destroys
it. The quality gate's entire purpose is to catch code that fails tests — and the failure path then
throws that code away, so the developer cannot inspect or fix it and must pay for the work again.

The inconsistency is notable: the **merge-failure** path deliberately preserves everything and says so
("Worktree preserved at: …"), while the **task-failure** path deletes it. The two paths disagree about
whether in-flight work is worth keeping.

**(2) Every run leaves a `workflow/*` branch that the supplied reaper cannot see.**

Three successive `smoke-test` runs on one project:

```
local  workflow/* : 3        remote workflow/* : 3
workflow/smoke-test-Ikka  ahead=0  sha=af3b80c
workflow/smoke-test-MhFg  ahead=0  sha=af3b80c
workflow/smoke-test-a09g  ahead=0  sha=af3b80c
git diff --stat master..workflow/smoke-test-a09g   ->   (empty)
```

All three are byte-identical to `master` — the workflow's only output is gitignored, so the
squash-merge replayed an empty patch. Each was still pushed to the remote and announced:

```
Integration branch workflow/smoke-test-a09g pushed to <origin>. Open a PR into master.
```

A PR from any of them would be empty. Branches are produced on the failure path too — a run that
completed **zero** tasks still pushed one (see #682's logs).

`dotbot prune-branches` cannot clean them with any documented invocation:

| Attempt | Result |
|---|---|
| `prune-branches --dry-run` | `✓ No branches match the prune criteria.` (window `30d`) |
| `prune-branches --dry-run --include-remote` | `✓ No branches match the prune criteria.` |
| `prune-branches --older-than 0d --dry-run` | `A parameter cannot be found that matches parameter name 'older-than'.` exit 0 |
| `prune-branches -OlderThan 0d --dry-run` | `✓ No branches match…` (skipped: they have upstreams) |
| `prune-branches -OlderThan 0d --include-remote --dry-run` | **3 candidates** ✓ |

Two defaults conspire. `Get-PrunableBranches:514` skips any branch with `has_remote_ref` unless
`-IncludeRemote`, and `_Get-RepoBranchRecords:566` derives that from `%(upstream:short)` — so **pushing
the branch is what makes it invisible** to the default prune. Widening the 30-day window then requires
`-OlderThan`, whose documented kebab-case form is unbindable (see BUG-006). Only the undocumented
PowerShell-native `-OlderThan` works.

A related correctness detail: `_Get-RepoBranchRecords:547` measures age from the branch tip's
`%(committerdate)`, not from when the branch was created. An empty integration branch inherits the
trunk's tip date, so on a repo whose trunk is older than 30 days brand-new branches are *immediately*
prunable, and on an actively committed trunk they never are. Branch age is not what is measured.

### What you expected

1. Work an agent has committed should survive a task failure — preserve the task branch (as the
   merge-failure path already does), or at minimum name the recoverable commit in the failure message.
   For a quality-gate failure specifically, keeping the branch is the whole point: someone has to fix
   the code.
2. A completed run should reap its own integration branch when it is empty, or the default
   `dotbot prune-branches` should be able to see the branches Dotbot itself creates — without requiring
   an undocumented parameter spelling plus a non-default flag.

### Steps to reproduce

**(1) Discarded work on a failed task** (needs an AI provider):

```powershell
$env:DOTBOT_HOME = '<dotbot checkout>'
# init a repo + dotbot init + `workflow add fast-prompt` as usual, then:
@'
{ "workflow": "fast-prompt",
  "quality_gate": { "enabled": true,
    "test_command": "pwsh -NoProfile -Command \"exit 3\"", "lint_command": null } }
'@ | Set-Content .bot\.control\settings.json

# launch a code task over HTTP (the CLI cannot pass a prompt):
& $env:DOTBOT_HOME\bin\dotbot.ps1 go        # note the port from .bot\.control\ui-port
Invoke-RestMethod -Method Post -Uri "http://localhost:<port>/api/workflows/fast-prompt/run" `
  -ContentType 'application/json' -Headers @{'X-Dotbot-Request'='1'} `
  -Body '{"prompt":"Create apps/web/src/lib/qg-probe.ts exporting QG_PROBE = \"x\". Nothing else."}'

# after it settles: task status = failed, and the committed file is on no branch
git for-each-ref --format='%(refname:short)' refs/heads/ |
  ForEach-Object { "$_ : $(git cat-file -e "${_}:apps/web/src/lib/qg-probe.ts" 2>$null; $LASTEXITCODE -eq 0)" }
git reflog --all | Select-String qg-probe        # -> nothing
git fsck --lost-found | Select-String 'dangling commit'   # -> the work is here only
```

**(2) Unreapable branch debris** (no AI provider needed):

```powershell
& $env:DOTBOT_HOME\bin\dotbot.ps1 workflow add smoke-test -y
1..3 | ForEach-Object { & $env:DOTBOT_HOME\bin\dotbot.ps1 run smoke-test --watch }

git branch --list 'workflow/*'                     # -> 3 branches
git branch -r --list 'origin/workflow/*'           # -> 3 on the remote too
git rev-list --count master..workflow/smoke-test-XXXX   # -> 0, they are empty

& $env:DOTBOT_HOME\bin\dotbot.ps1 prune-branches --dry-run                    # -> none
& $env:DOTBOT_HOME\bin\dotbot.ps1 prune-branches --older-than 0d --dry-run    # -> binding error, exit 0
& $env:DOTBOT_HOME\bin\dotbot.ps1 prune-branches -OlderThan 0d --include-remote --dry-run   # -> finally lists them
```

### Environment

```
OS: Windows 11 Pro 10.0.26200 | dotbot v4.0.2 (main @ 7c95b466)
pwsh 7.6.4 | git 2.54.0.windows.1 | claude CLI 2.1.222 (models: shipped defaults, all resolve)
DOTBOT_HOME set explicitly; %APPDATA%\dotbot\user-settings.json absent
Target repo: 1129 files, single remote (local bare origin), master default
```

### Severity

high

### Logs / screenshots

Cosmetic defect visible in the prune output that finally works — branch name and timestamp are
concatenated with no separator:

```
  CANDIDATES
  ────────────────────────────────────────────
      • workflow/smoke-test-Ikka08/05/2026 15:37:47 [has remote]
      • workflow/smoke-test-MhFg08/05/2026 15:37:47 [has remote]
      • workflow/smoke-test-a09g08/05/2026 15:37:47 [has remote]
```

Reap command that does work today, for anyone accumulating debris now:

```powershell
dotbot prune-branches -OlderThan 0d --include-remote --dry-run   # review first
dotbot prune-branches -OlderThan 0d --include-remote -y
```

### LINKED ISSUE #684: CLI: documented flags are unbindable or silently ignored, and no error path exits non-zero
### What happened

A sweep of every verb across three contexts (non-git directory, git repo without `.bot`, initialised
project — 168 invocations) found three distinct flag-handling failure modes. **All of them exit 0.**

**Mode 1 — real flags that fail with a raw PowerShell error.**

`ConvertTo-SplatArg` (`bin\dotbot.ps1:182-190`) strips leading dashes and uses the flag text
**verbatim** as the splat key, so `--older-than` becomes the key `older-than`. Binding then requires a
parameter or alias literally named `older-than`. Whether one exists is per-script and inconsistent —
in the same param block:

```powershell
# prune-branches.ps1:22-28
[string]$OlderThan = '30d',                        # <-- no alias  => --older-than FAILS
[ValidateSet('workflow','task','all')]
[string]$Match = 'all',                            #     no hyphen => --match works
[Alias('dry-run')][switch]$DryRun,                 # <-- alias     => --dry-run works
[Alias('include-remote')][switch]$IncludeRemote,   # <-- alias     => works
```

Confirmed broken, each reproduced twice:

| Command | Error | Exit |
|---|---|---|
| `dotbot prune-branches --older-than 90d --dry-run` | `A parameter cannot be found that matches parameter name 'older-than'.` | 0 |
| `dotbot logs --poll-interval-ms 500 --tail 2` | `A parameter cannot be found that matches parameter name 'poll-interval-ms'.` | 0 |
| `dotbot runtime-status --json` | `A parameter cannot be found that matches parameter name 'json'.` | 0 |

`--older-than` matters beyond cosmetics: it is the only control over `prune-branches`' 30-day window,
so run debris cannot be cleaned through the documented CLI at all (see #683).

What the operator sees is a stack frame inside dotbot's own dispatcher:

```
prune-branches.ps1: C:\projects\dotbot\dotbot\bin\dotbot.ps1:747
Line |
 747 |  … anches" { & (Join-Path $ScriptsDir 'prune-branches.ps1') @SplatArgs }
     |                                                             ~~~~~~~~~~
     | A parameter cannot be found that matches parameter name 'older-than'.
```

**Mode 2 — flags silently ignored, producing a false result.**

`doctor.ps1:14` declares a bare `param()` with no `[CmdletBinding()]` and no `[Parameter()]`
attribute, making it a *simple* script: unrecognised named arguments land in `$args` and are discarded
rather than rejected. So the kebab-case form of its own documented parameter is dropped, and doctor
scans the default `./.bot` while reporting on a path the operator never named:

```
> dotbot doctor --bot-root C:\does\not\exist
    Project: tanda                     <- scanned the wrong directory
  PASS: 8    WARN: 0    FAIL: 0
  exit=0

> dotbot doctor -BotRoot C:\does\not\exist
    Project: not
  ✗ .bot directory not found at: C:\does\not\exist
  exit=2                               <- correct, via the PowerShell-native spelling
```

A health check that reports PASS for a directory it did not examine is worse than one that errors.
`doctor --json` is swallowed the same way, so a caller asking for JSON receives ANSI-decorated human
text with exit 0. (`--json` in fact exists only on `status.ps1`.)

**Mode 3 — no error path sets a non-zero exit code.** Every one of these exits 0:

| Invocation | Output |
|---|---|
| `dotbot bogus-verb` | `✗ Unknown command: bogus-verb` |
| `dotbot run` (no name) | `⚠ Usage: dotbot run <workflow-name> …` |
| `dotbot workflow run` (no name) | usage |
| `dotbot run smoke-test --wathc` (typo) | `A parameter cannot be found that matches parameter name 'wathc'.` |
| `dotbot run smoke-test --watch --poll-interval-ms 100` | `ValidateRange` violation (min 250) |
| `dotbot logs --tail 0` | `ValidateRange` violation (min 1) |
| `dotbot prune-branches --match bogus` | `ValidateSet` violation |
| `dotbot install content` | usage |

`bin\dotbot.ps1:753-758` prints `Unknown command` and falls out of the `switch` with no `exit 1`;
`Invoke-Run:539` and the `workflow run` path do the same for a missing name. Only `doctor` (`:740-742`)
and `run` with a *valid* workflow (`:535-537`) propagate a child exit code. Consequently
`dotbot prune-branches --older-than 7d && echo ok` prints `ok` having pruned nothing, and no CI step
can detect a mistyped dotbot command.

Also confirmed in the sweep:

- **`dotbot help` advertises `install content`**, which `Invoke-Install:505` does not accept (it takes
  `runtime | agent(s) | prompt(s) | skill(s)`). Typing it prints usage, exit 0.
- **`dotbot help` omits `workflow scaffold`**, which exists and works — and whose usage line advertises
  `--Force` in PascalCase, unlike every other documented flag.
- **`dotbot tasks run` silently drops all flags**: `Invoke-Tasks:546-549` calls `& $script` with no
  splat at all, so `dotbot tasks run --json` is accepted and ignored.
- **`dotbot studio` is unreachable**: `:702` resolves `$DotbotBase\studio-ui` but the directory is
  `src\studio-ui`, so it always prints `✗ Studio not found.` — and its remedy, `⚠ Run 'dotbot update'
  to install the studio`, cannot help, because `dotbot update` only prints `git pull`.
- `profiles` is an undocumented alias for `list`.

### What you expected

1. A documented flag should bind. Either add the kebab-case `[Alias()]` on every multi-word parameter
   (`OlderThan`, `PollIntervalMs`, `BotRoot`, …) or normalise `ConvertTo-SplatArg` to map
   `--older-than` → `OlderThan` once, centrally.
2. An unrecognised flag should be **rejected**, never silently dropped — `doctor.ps1` should be an
   advanced script like its siblings.
3. Every error path should exit non-zero: unknown verb, missing required argument, parameter-binding
   failure, validation failure. Errors should also be reported through the theme helpers rather than
   surfacing raw PowerShell stack frames.
4. `dotbot help` should match reality: drop `install content`, add `workflow scaffold`.

### Steps to reproduce

```powershell
$env:DOTBOT_HOME = '<dotbot checkout>'
$db = "$env:DOTBOT_HOME\bin\dotbot.ps1"
cd <any initialised dotbot project>

# Mode 1 — unbindable documented flags, all exit 0
& $db prune-branches --older-than 90d --dry-run ; "exit=$LASTEXITCODE"
& $db logs --poll-interval-ms 500 --tail 2      ; "exit=$LASTEXITCODE"
& $db runtime-status --json                     ; "exit=$LASTEXITCODE"

# Mode 2 — silently ignored, false PASS on a directory that does not exist
& $db doctor --bot-root C:\does\not\exist       ; "exit=$LASTEXITCODE"   # PASS 8, exit 0
& $db doctor -BotRoot   C:\does\not\exist       ; "exit=$LASTEXITCODE"   # correct failure, exit 2

# Mode 3 — no non-zero exit anywhere
& $db bogus-verb                                ; "exit=$LASTEXITCODE"
& $db run                                       ; "exit=$LASTEXITCODE"
& $db run smoke-test --wathc                    ; "exit=$LASTEXITCODE"
& $db logs --tail 0                             ; "exit=$LASTEXITCODE"
& $db install content                           ; "exit=$LASTEXITCODE"
& $db studio                                    ; "exit=$LASTEXITCODE"
```

### Environment

```
OS: Windows 11 Pro 10.0.26200 | dotbot v4.0.2 (main @ 7c95b466)
pwsh 7.6.4 | git 2.54.0.windows.1
DOTBOT_HOME set explicitly; %APPDATA%\dotbot\user-settings.json absent
Contexts swept: non-git directory; git repo without .bot; initialised project (1129-file repo)
```

### Severity

medium

### Logs / screenshots

Sweep totals, identical in the non-git and no-`.bot` contexts and near-identical in a live project:

| Context | Invocations | exit≠0 | Raw PS errors | exit 0 with `Usage:` | exit 0 with `Unknown command` |
|---|---|---|---|---|---|
| non-git | 57 | 4 | 8 | 13 | 2 |
| git, no `.bot` | 57 | 4 | 8 | 13 | 2 |
| initialised project | 54 | 1 | 8 | 13 | 2 |

Full per-invocation records (args, exit code, duration, output) are in the attached CSVs; complete
untruncated output per invocation is in the matching `.txt` files.

Every multi-word CLI parameter and whether it carries the kebab alias the dispatcher needs:

```
MISSING the alias (--flag hard-fails):
  prune-branches.ps1  OlderThan        --older-than
  logs.ps1            PollIntervalMs   --poll-interval-ms
  doctor.ps1          BotRoot          --bot-root          (silently ignored instead: lax param())
  serve.ps1           MothershipApiKey --mothership-api-key   (has 'mothership-key' only)
  install-content.ps1 GlobalInstall    --global-install       (has 'global' only)

HAS the alias (works):
  init-project.ps1    CopyRuntime      --copy-runtime
  init-project.ps1    DryRun           --dry-run
  prune-branches.ps1  DryRun           --dry-run
  prune-branches.ps1  IncludeRemote    --include-remote
```

### LINKED ISSUE #685: Dashboard loads webfonts from fonts.googleapis.com; degrades on offline / egress-filtered networks
### What happened

Driving the dashboard in Google Chrome stable, two font requests fail:

```
GET https://fonts.gstatic.com/s/inter/v20/UcC73FwrK3iLTeHuS_nVMrMxCp50SjIa1ZL7.woff2
    — net::ERR_FAILED
GET https://fonts.gstatic.com/s/jetbrainsmono/v24/tDbv2o-flEEny0FZhsfKu5WU4zr3E_BX0PnT8RD8yKwBNntkaToggR7BYRbKPxDcwg.woff2
    — net::ERR_FAILED
```

Source — `src\ui\static\css\base.css:5`:

```css
@import url('https://fonts.googleapis.com/css2?family=JetBrains+Mono:wght@400;500;600;700&family=Inter:wght@400;500;600&display=swap');
```

So a dashboard that otherwise runs entirely on localhost makes an outbound request to
`fonts.googleapis.com` (and then `fonts.gstatic.com`) on every page load. Two consequences:

1. **Offline / air-gapped / egress-filtered environments.** The requests fail and the CRT-style
   typography the design depends on silently falls back to system fonts. The dashboard remains usable —
   this is not a functional break — but it does not look as intended, and there is no indication why.
2. **An unnecessary third-party dependency for a local developer tool.** Every dashboard load tells
   Google that someone opened it. For a tool whose stated model is local-first and whose own docs
   emphasise an auditable local trail, that is surprising.

`JetBrains Mono` and `Inter` are both SIL Open Font License, so they can be vendored without licensing
friction.

### What you expected

The dashboard renders identically with no network access. Self-host the two families under
`src/ui/static/` (subset woff2, `@font-face` with `font-display: swap`) and drop the remote `@import`,
keeping a system-font stack as the fallback.

If a remote fetch is deliberate, it should at least be degradable on purpose — a local `@font-face`
fallback so the intended typography survives without egress.

### Steps to reproduce

```powershell
$env:DOTBOT_HOME = '<dotbot checkout>'
cd <an initialised dotbot project>
& $env:DOTBOT_HOME\bin\dotbot.ps1 go
# read the port from .bot\.control\ui-port, then open http://localhost:<port>
```

In Chrome DevTools → Network, filter `gstatic`: two woff2 requests are issued on load. To see the
failure mode directly, block `fonts.googleapis.com` and `fonts.gstatic.com` (DevTools → Network request
blocking, or an offline host) and reload — the requests fail and the fonts fall back.

Scripted, with Playwright against Chrome stable:

```ts
const failed: string[] = [];
page.on("requestfailed", r => failed.push(`${r.method()} ${r.url()} — ${r.failure()?.errorText}`));
await page.goto("/", { waitUntil: "domcontentloaded" });
// -> failed contains the two fonts.gstatic.com woff2 requests
```

Or with no browser at all:

```powershell
Select-String -Path "$env:DOTBOT_HOME\src\ui\static\css\base.css" -Pattern 'fonts.googleapis.com'
```

### Environment

```
OS: Windows 11 Pro 10.0.26200 | dotbot v4.0.2 (main @ 7c95b466)
pwsh 7.6.4 | Google Chrome stable (channel: "chrome" via Playwright 1.56.1)
Harness: standalone Playwright project outside DOTBOT_HOME; 4 of 5 checks passed
Network: outbound to fonts.gstatic.com blocked in this environment
```

### Severity

low

### Logs / screenshots

The rest of the dashboard checks passed in the same run, so this is isolated to the font fetch:

```
✓ /api/info and /api/state return 200 with a non-empty body
✓ CSRF guard refuses a state-changing POST without X-Dotbot-Request (403 "Missing CSRF header")
✓ upload filename sanitiser resists ../ traversal, backslash traversal, CON.md, invalid chars
✓ a second browser tab attaches without errors
✗ page load reports 2 failed requests — both fonts.gstatic.com woff2

Page title: "DOTBOT // Control Panel"
No uncaught JavaScript errors; no HTTP >= 400 from the dotbot server itself.
```

### LINKED ISSUE #686: Expose `git.base_branch` in the dashboard Settings tab
### Problem statement

`git.base_branch` is the only setting that controls which branch Dotbot cuts task worktrees from and
squash-merges into. Verified end to end: setting it redirects merges away from the trunk, so it is the
one available answer for a team working on a long-lived branch instead of `master`.

```json
// .bot/.control/settings.json
{ "git": { "base_branch": "feature/qa-work" } }
```

```
New-TaskWorktree      -> worktree-map base_branch = "feature/qa-work"
Complete-TaskWorktree -> "Squash-merged to feature/qa-work and cleaned up"
master unchanged; feature/qa-work advanced
```

But there is no way to set it other than hand-editing JSON:

- `dotbot help` lists no command for it; there is no `dotbot config` verb at all.
- `src\ui\modules\SettingsAPI.psm1:447-477` already implements `Get-GitConfig` and `Set-GitConfig`, and
  exports both at `:1355-1356` — but **no HTTP route in `server.ps1` calls them**, and the Settings tab
  renders no control for them. A grep for `settings/git` across `src\ui` returns nothing.
- The file it lives in, `.bot/.control/settings.json`, is gitignored, so the value cannot be committed
  and shared. Every developer must recreate it on every machine, by hand, with no validation and no
  discoverable documentation.

The practical result is that a team hitting the wrong-merge-target behaviour (#679) has a working fix
available and no way to find it. Nothing in the failure message, the `needs-input` escalation, or
`dotbot doctor` mentions that this setting exists.

### Proposed outcome

A **Base branch** field in the dashboard's Settings tab, under a Git section:

- Populated from `Get-GitConfig`; saved through `Set-GitConfig` (both already written).
- Validated against `git branch --format='%(refname:short)'` so a typo cannot be saved — currently an
  invalid value makes `Resolve-DotbotBaseBranch:172` throw at run time
  (`git.base_branch is set to 'x' but no such branch exists … refusing to fall back to main/master`),
  which surfaces late and confusingly.
- Empty input stores `null`, preserving today's `main` → `master` probe. `SettingsAPI.psm1:452-465`
  already normalises empty string to `null`.
- Shown with the currently resolved value when unset (e.g. *"unset — resolving to `master`"*), so the
  operator can see what Dotbot will actually do.

A `dotbot config get|set git.base_branch <branch>` verb would cover the headless/CI case, but the UI
field alone would resolve the discoverability problem.

Worth pairing with the fix for #680: today the setting is honoured on the happy path but abandoned by
the three failure-path `Assert-OnBaseBranch` calls, so exposing it in the UI without that fix would
promise more consistency than it delivers.

### Affected users / use case

Any team whose integration branch is not `main` or `master` — release branches, `develop`, or a long-lived feature branch shared by several people. This is the reported configuration of at least one team hitting #679, and it is the only supported way to stop Dotbot writing to the trunk.

### Rough size

S

### Additional context, links, mockups

The API layer is already done; this is one HTTP route plus one input control. For reference, the value
resolves correctly only when a `BotRoot` is supplied:

```
Resolve-DotbotBaseBranch -ProjectRoot <r> -BotRoot <r>\.bot   ->  feature/qa-work
Resolve-DotbotBaseBranch -ProjectRoot <r>                     ->  master
```

Settings layers, low → high (`Dotbot.Settings.psm1:130-135`), so `.bot/.control/settings.json` wins:

```
<DOTBOT_HOME>\content\settings\settings.default.json
<BotRoot>\content\settings\settings.default.json
%APPDATA%\dotbot\user-settings.json
<BotRoot>\.control\settings.json        <- highest, gitignored
```

A tracked alternative already exists — `<BotRoot>\content\settings\settings.default.json` is committed —
so documenting *that* path as the place to set a shared base branch would help teams immediately, even
before any UI work.

EnmaJim:
## Linked issue

Closes #679
Closes #680
Closes #681
Closes #682
Closes #683
Closes #684
Closes #685
Closes #686

## Summary of changes

A reliability pass over the CLI, the runtime lifecycle and task worktrees. Eight issues
that shared the same root cause \Uffffffff paths that fail silently and errors that never reach
the shell \Uffffffff fixed on one branch.

**CLI dispatch and exit codes (#684)**

- `bin/dotbot.ps1` used the raw flag text as the splat key, so every hyphenated flag
  (`--older-than`, `--poll-interval-ms`, `--bot-root`, `--mothership-api-key`,
  `--no-auto-runtime`, `--no-runtime`) raised a `ParameterBindingException` from inside
  the dispatcher. A single `ConvertTo-DotbotParameterName` helper now normalises
  `--older-than` ? `OlderThan` at the five places the dispatcher derives a splat key.
- Every error path exited 0: unknown verb, missing required argument, parameter-binding
  failure, and `ValidateSet`/`ValidateRange` violations. They now exit 1 and are
  reported through the theme helpers instead of a raw PowerShell stack frame. `init`,
  `workflow`, `registry`, `install`, `tasks`, `status`, `serve`, `prune-branches`,
  `studio`, `logs` and `runtime-status` propagate their child script's exit code.
- `doctor.ps1`, `tasks-run.ps1` and `tasks-stop.ps1` are now advanced scripts that
  reject unknown flags \Uffffffff `doctor --bot-root <path>` used to report PASS for a directory
  it never scanned.
- `dotbot studio` is reachable again (it resolved `<DOTBOT_HOME>/studio-ui`, but the
  studio moved to `src/studio-ui/` in v4), and `dotbot help` now matches the real
  commands (`install skill`, `workflow scaffold`, `--force`).

**Runtime lifecycle (#682)**

- The runtime precondition and the auto-start were both nested inside the `--watch`
  branch, so the default path spawned the detached task runner with no runtime and
  exited 0; the runner then parked every task in `needs-input` asynchronously, in a log
  file the caller never sees, and only after an integration branch had been created and
  pushed. `dotbot run` and `dotbot tasks run` now start the runtime themselves, and
  `--no-auto-runtime` refuses synchronously with exit 1 _before_
  `Initialize-WorkflowRun`, so a refused run leaves nothing behind.
- New `Start-DotbotRuntimeDetached`: brings the runtime up in a process of its own by
  spawning `dotbot serve` and waiting for the connection file, which
  `Start-DotbotRuntime` writes only once the listener is accepting. The result carries
  no listener handle, so a caller can never tear down a runtime it does not own.
- `Test-RuntimeAlive` no longer trusts a recorded PID on its own: it also requires the
  process to be a PowerShell host, and the new `Test-RuntimeServing` probes the recorded
  URL before reusing it. A stale connection file \Uffffffff the normal state after any
  non-graceful exit \Uffffffff made `dotbot run` print `Using existing headless runtime.` and
  then do nothing.
- Concurrent cold starts are serialised on a named mutex keyed on the bot root, the
  readiness timeout kills the child it spawned, and `Stop-DotbotRuntime` removes the
  connection file only when it names the process doing the stopping. A new
  `dotbot runtime stop` verb stops a detached runtime that has no console to close.

**Worktrees and task merges (#679, #680, #681, #683)**

- `Complete-TaskWorktree` took its target from the `base_branch` that
  `worktree-map.json` recorded when the worktree was created, never reconciling it with
  the branch the main checkout was on: with a clean tree it silently squash-merged into
  the trunk; with a dirty file under `.bot/workspace/decisions/` it failed and parked the
  run. The target is now resolved by precedence (explicit `-BaseBranch` ? configured
  `git.base_branch` ? the recorded value when it matches the checkout ? the checked-out
  branch), reconciled back into the map, and named in the result message. `task/*`
  branches and detached HEADs are never adopted.
- `Assert-OnBaseBranch` reports git's stderr (previously every blocker produced the same
  unactionable message) and takes `-BotRoot`, so its fallback no longer yanks the working
  copy off a configured base branch on any failed or skipped task.
- `Ensure-DotbotWorktreeExcludes` wrote to `.git/info/exclude`, which git keeps in the
  shared common directory, so its entries silently ignored those _tracked_ trees in the
  main checkout: `git add .bot/workspace/tasks/` became a no-op and run state accumulated
  invisibly. Suppression now uses a nested `.gitignore` inside each generated copy.
  Already-affected repositories self-heal on their next task.
- A failed task no longer discards the commits the agent already made: all three failure
  paths in `Invoke-WorkflowProcess.ps1` removed the worktree _and_ the branch, leaving
  committed work reachable only as a dangling object until the next `git gc`. They now
  remove only the worktree and report the preserved branch with its tip SHA.
- A run no longer leaves an empty, unreapable `workflow/*` integration branch: run
  completion counts `rev-list --count <base>..<integration>` and, when it is zero, skips
  the push and deletes the branch with the safe `-d` \Uffffffff unless a retained worktree still
  records it as its base.

**Paths containing brackets**

- PowerShell reads `[...]` in `-Path` as a character class, so in a directory such as
  `bracket[v2]` every composed probe missed: `Get-DotbotProjectBotPath` returned the
  temp-directory fallback and `dotbot init` reported "not a git repository" for a repo
  that was one. Every `Test-Path (Join-Path \Uffffffff)` in `src/`, `bin/` and `content/` now
  passes `-LiteralPath`, as do the content reads and writes and the stale-worktree
  cleanup.

**Dashboard (#685)**

- `src/ui/static/css/base.css` `@import`ed Google Fonts, so a tool that otherwise runs
  entirely on localhost made an outbound request on every page load and silently fell
  back to system fonts offline. Inter and JetBrains Mono (both OFL-1.1) are now
  self-hosted as variable woff2 under `src/ui/static/fonts/`, matching the
  `unicode-range` splits (latin and latin-ext) the CDN stylesheet used, and the server
  serves `.woff2`/`.woff` with the correct MIME type.

**`git.base_branch` discoverability (#686)**

- `git.base_branch` is the only setting that decides which branch a run is cut from and
  merged into, and nothing surfaced it: `dotbot help` listed no command for it, and a
  team hitting the wrong-merge-target behaviour of #679 had a working fix available and
  no way to find it. `dotbot help` now names the setting alongside the file it lives in
  (`.bot/.control/settings.json`), and the README documents it with an example, the
  fail-fast behaviour when the configured branch does not exist, and the warning a run
  prints when the configured base differs from the checked-out branch.
- Not in this change: the HTTP route wiring the existing `Get-GitConfig`/`Set-GitConfig`
  in `src/ui/modules/SettingsAPI.psm1` to a control in the dashboard Settings tab. The
  setting is now documented and discoverable, but still edited as JSON.

## Testing notes

```powershell
./tests/Run-Tests.ps1              # layers 1-3
./tests/Run-Tests.ps1 -Layer 1     # static gates, including Test-NoWildcardPaths
```

Tests added or extended on this branch:

| File                              | Covers                                                                      |
| --------------------------------- | --------------------------------------------------------------------------- |
| `tests/Test-CliDispatch.ps1`      | kebab-case flag binding and a non-zero exit code on every error path (#684) |
| `tests/Test-ProcessDispatch.ps1`  | `--poll-interval-ms` and the CLI RUNTIME PRECONDITION section (#682)        |
| `tests/Test-NoWildcardPaths.ps1`  | layer 1 gate: zero composed probes without `-LiteralPath`, no allowlist     |
| `tests/Test-Components.ps1`       | UI components                                                               |
| `tests/Test-Runtime.ps1`          | `Test-RuntimeAlive` / `Test-RuntimeServing` and the detached start          |
| `tests/Test-WorkflowManifest.ps1` | workflow manifest                                                           |

`Run-Tests.ps1` also tees each child's output and reports a file that exits non-zero
**without** printing a `Summary:` line as a distinct crash \Uffffffff which is how a suite that
died mid-run went unnoticed before.

Suggested manual verification:

1. `dotbot run <workflow>` without `--watch` from a feature branch ? the runtime starts
   itself, the run progresses, and you end up back on your branch (not `main`).
2. `dotbot tasks run --no-auto-runtime` with no runtime ? an error message and
   `$LASTEXITCODE` of 1, with no integration branch created.
3. `dotbot doctor --bot-root <path>` ? actually scans that path.
4. `dotbot prune-branches --older-than 7` ? binds instead of raising
   `ParameterBindingException`.
5. Clone into a directory with brackets (`bracket[v2]`) and run `dotbot init`.
6. `dotbot serve` with the network cut ? the dashboard keeps its typeface.
7. `dotbot help` ? lists `git.base_branch` and the settings file it belongs in.

## Checklist

- [x] Tests added or updated
- [x] Docs updated (if behaviour changed)
- [x] Linked issue exists


EnmaJim:
@carlospedreira please review this PR

carlospedreira (approved):

